PR #5, opened 2021-01-19
Rename caseInsensitive to caseSensitive

lukasmalkmus:


bahlo (approved):

lukasmalkmus:
Let me fix the commit name. Something went wrong here.

axiom/monitors_test.go
```diff
@@ -298,7 +298,7 @@ func TestMonitor_MarshalJSON(t *testing.T) {
 				"op": "",
 				"field": "",
 				"value": null,
-				"caseInsensitive": false,
+				"caseSensitive": false,
```
bahlo, line 301:
```suggestion
				"caseSensitive": true,
```
Shouldn't this be flipped?

lukasmalkmus: No. The test payload which gets encoded does not explicitly specify `monitor.CaseSensitive = true`. The zero value of `bool` is still false.

bahlo: Didn't look at the test, sounds good 👍

bahlo (approved):

PR #8, opened 2021-02-16
Support for organization id

lukasmalkmus:
Adds support for the organization id. Since there already is a constructor for cloud-only deployments, I just added it to the "default" constructor.

### DIFF
--- a/axiom/client.go
+++ b/axiom/client.go
@@ -79,6 +79,7 @@ type Client struct {
 	baseURL        *url.URL
 	userAgent      string
 	accessToken    string
+	orgID          string
 	strictDecoding bool
 
 	httpClient *http.Client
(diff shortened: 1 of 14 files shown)


axiom/client.go
```diff
@@ -136,9 +139,9 @@ func NewClient(baseURL, accessToken string, options ...Option) (*Client, error)
 }
 
 // NewCloudClient is like NewClient but assumes the official Axiom Cloud url as
-// base url.
+// base url and does not accept an organization ID.
 func NewCloudClient(accessToken string, options ...Option) (*Client, error) {
```
bahlo, line 143: This is the wrong way around. Org-IDs are required for Cloud, not for self-host

axiom/client_test.go
```diff
@@ -20,6 +20,8 @@ const (
 	endpoint = "http://axiom.local"
 	// accessToken is a placeholder access token.
 	accessToken = "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"
+	// orgID is a placeholder organization token.
+	orgID = "YYYYYYYY-YYYY-YYYY-YYYY-YYYYYYYYYYYY"
```
bahlo, line 24: OrgID is not a UUID

bahlo (changes requested):
OrgID should be in the cloud constructor

bahlo (approved):

PR #10, opened 2021-03-01
Trim support

lukasmalkmus:
Simple one. Just hitting an existing endpoint.

I'm pretty sure, CI will fail. We need a server-side fix and wait for that to hit our staging env. Or we switch over to the dev env 
for integration tests. WDYT @bahlo?


axiom/datasets.go
```diff
@@ -281,6 +287,26 @@ func (s *DatasetsService) Info(ctx context.Context, id string) (*DatasetInfo, er
 	return &res, nil
 }
 
+// Trim the dataset identified by its id to a given length. The max duration
+// given will mark the oldest timestamp an event can have. Older ones will be
+// deleted from the dataset.
+func (s *DatasetsService) Trim(ctx context.Context, id string, maxDuration time.Duration) (*TrimResult, error) {
+	req := struct {
```
bahlo, line 294: Can we make this a proper struct? Can be internal obv

axiom/version.go
```diff
@@ -18,7 +13,10 @@ type VersionService service
 
 // Get the version of a deployment.
 func (s *VersionService) Get(ctx context.Context) (string, error) {
-	var res version
+	var res struct {
+		CurrentVersion string `json:"currentVersion"`
+	}
```
bahlo, line 18: Same with external struct

axiom/datasets.go
```diff
@@ -281,6 +287,26 @@ func (s *DatasetsService) Info(ctx context.Context, id string) (*DatasetInfo, er
...
+		MaxDuration time.Duration `json:"maxDuration"`
```
bahlo, line 295: I thought the format was string?

axiom/datasets.go
lukasmalkmus: Why? I think this is a nice way of keeping some internals from that function out of the package scope.

lukasmalkmus: On the other hand: This will portably extend in the future so you are kinda right.

axiom/datasets.go
lukasmalkmus: Yep yep, already fixed that locally :)

mhr3 (approved):

PR #20, opened 2021-03-11
Basic Organization support

lukasmalkmus:
Basic support for Organizations. This does not (yet) cover the cloud-only endpoints for billing/usage/etc.

Also fixed a typo along the way.


Makefile
```diff
@@ -73,7 +73,13 @@ fmt: ## Format and simplify the source code using `gofmt`
 	@! $(GOFMT) -s -w $(shell find . -path -prune -o -name '*.go' -print) | grep '^'
 
 .PHONY: generate
-generate: $(STRINGER) axiom/datasets_string.go axiom/monitors_string.go axiom/notifiers_string.go axiom/starred_string.go axiom/users_string.go ## Generate code using `go generate`
+generate: $(STRINGER) \
```
lukasmalkmus, line 76: Here is a special one for you @bahlo: This change breaks `make help`. The comment needs to be on the same line as the recipe. However, moving it up, breaks `make`. I think you introduced the improved `help` command a while back in on of our projects. So maybe you have a nifty idea in mind :) On the other hand, this is not a blocker.

bahlo (approved):
LGTM

PR #29, opened 2021-03-18
Cleanup Makefile and README

lukasmalkmus:
I removed a tool dependency which was unused and thus some lines in the `Makefile` as well.

While at it, I also removed two lines from the `README` I think aren't really useful: Go developers have Go installed and this is a Go library. We don't need to tell them to install. Furthermore `go mod` has become the de-facto standard for dependency management. No need to talk about that in the `README`, either.


thecraftman (approved):
LGTM

PR #35, opened 2021-04-28
Enable http/2 transport

lukasmalkmus:
The transport needs to be explicitly configured to use HTTP/2.


mhr3 (changes requested):
I think this is all you need

https://golang.org/pkg/net/http/#Transport.ForceAttemptHTTP2

mhr3 (dismissed):
👍🏼 

But maybe `go mod tidy`

lukasmalkmus:
Doing so works, but `make dep` will alter the `go.sum` file again and add back the entries. There must be a reason for this... But committing after `go mod tidy` breaks the `check` stage.

mhr3 (approved):

PR #36, opened 2021-05-12
Handle ingest-only token for non-ingest endpoints

lukasmalkmus:
* Returns an error, if an ingest-only token is used for non-ingest operations
* Adds two new, public helper methods for identifying tokens
* Adds a new client option for setting the access token.

### DIFF
--- a/axiom/client.go
+++ b/axiom/client.go
@@ -17,8 +17,14 @@ import (
 // CloudURL is the url of the cloud hosted version of Axiom.
 const CloudURL = "https://cloud.axiom.co"
 
-// ErrUnauthenticated is raised when the access token isn't valid.
-var ErrUnauthenticated = errors.New("invalid authentication credentials")
+var (
+	// ErrUnauthenticated is raised when the access token isn't valid.
+	ErrUnauthenticated = errors.New("invalid authentication credentials")
+
+	// ErrUnprivilegedToken is raised when a client tries to call a non-ingest
+	// endpoint with an ingest-only token configured.
+	ErrUnprivilegedToken = errors.New("using ingest token for non-ingest operation")
+)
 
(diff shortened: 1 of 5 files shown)


axiom/client.go
```diff
@@ -198,8 +212,12 @@ func (c *Client) call(ctx context.Context, method, path string, body, v interfac
 // newRequest creates an API request. If specified, the value pointed to by
 // body will be included as the request body. If it isn't an io.Reader, it will
 // be included as a JSON encoded request body.
-func (c *Client) newRequest(ctx context.Context, method, endpoint string, body interface{}) (*http.Request, error) {
-	rel, err := url.ParseRequestURI(endpoint)
+func (c *Client) newRequest(ctx context.Context, method, path string, body interface{}) (*http.Request, error) {
+	if IsIngestToken(c.accessToken) && path != "/api/v1/tokens/ingest" {
```
mhr3, line 216: I assume we don't call newRequest() when doing actual ingest? Otherwise ingest will just return ErrUnprivilegedToken

lukasmalkmus: Ups, good catch. This API path here should actually be the ingest path (not the path for ingest-only tokens). Will add some more tests and fix this.

axiom/client_test.go
```diff
@@ -231,6 +260,47 @@ func TestClient_do_RedirectLoop(t *testing.T) {
...
+		{
+			input: "/api/v1/datasets/test/",
+			match: false,
+		},
+		{
+			input: "/api/v1/tokens/personal/validate",
+			match: false,
+		},
+		{
+			input: "/api/v1/tokens/validate",
```
mhr3, line 293: shouldn't this one be allowed as well?

lukasmalkmus: True!

mhr3 (approved):

PR #42, opened 2021-06-01
Reorganize workflows

lukasmalkmus:
This reorganizes the single workflow into multiple ones. It also gets rid of a lot of blank lines, self explaining descriptions, etc. Also improves caching performance.


.github/workflows/pr.yml
```diff
@@ -0,0 +1,69 @@
+name: PR
+
+on:
+  pull_request:
+    branches:
+      - main
```
bahlo, line 6: Can we run this workflow  on master as well to catch semantic merge conflicts? And maybe releases as well just to be safe.

bahlo (changes requested):
Small question

.github/workflows/pr.yml
lukasmalkmus: We could. But the repo settings just allow rebase commits so the HEAD of your PR branch represents the HEAD of the main branch, if the PR lands.

bahlo: TBH even with the checks I'd feel way better if the tests ran on `main` and tags.

lukasmalkmus: Well, I'd say, only `main`. Because when we tag a release, we do that on a main branch that passed already, right?

bahlo (approved):

PR #53, opened 2021-08-18
Improved testing

lukasmalkmus:
* Support the latest two Go releases and properly test against them
* Only update coverage and and code quality report from tests that ran the latest Go release
* Introduce `AXIOM_DATASET_PREFIX` as in `axiom-node` to uniquely identify a test dataset if it should be left over
* Skip Go installation for the `lint` action
* Use `github.token` instead of `secrets.GITHUB_TOKEN`. The reason for this is that to unexperienced users it looks like `GITHUB_TOKEN` is configured as a repository secret like all the other secretes for the `secrets.*` key. This makes it a bit more implicit, where the token comes from.
* Run regression test at night
* Some README fixes

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -5,18 +5,20 @@ on:
     branches:
       - main
 
-env:
-  GOVERSION: "1.16"
-
 jobs:
   gen-diff:
     name: Codegen diff
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        go:
+          - 1.16
+          - 1.17
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
         with:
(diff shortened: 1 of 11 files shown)


mhr3:
> Also bump to latest Go 1.17
Bad @lukasmalkmus, you don't do these things literally a day before release

lukasmalkmus:
> Bad @lukasmalkmus, you don't do these things literally a day before release

@mhr3 I didn't change module requirements, bump dependencies or build any binary with Go 1.17 (since this is a package with nothing to build). I just test against the latest Go release, which should be just fine.

bahlo (approved):

mhr3:
> I didn't change module requirements, bump dependencies or build any binary with Go 1.17

Ah cool, didn't look at it properly

PR #61, opened 2021-09-03
Validate credentials

lukasmalkmus:
* New `ValidateCredentials` method to let users validate the client configuration before making calls or knowing any internals of which methods to use for validating different types of tokens
* Improve regex that only matches ingest-token enabled endpoints

### DIFF
--- a/axiom/axiom.go
+++ b/axiom/axiom.go
@@ -1,6 +1,9 @@
 package axiom
 
-import "strings"
+import (
+	"context"
+	"strings"
+)
 
 // IsIngestToken returns true if the given acces token is an ingest token. If
 // false is returned, that does not imply that the token is a personal token.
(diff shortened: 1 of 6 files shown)


lukasmalkmus:
@mhr3 Could you do me a favour and double check the regex, I tweaked? The idea is to more explicitly allow query parameters on the `ingest` endpoint.

mhr3:
> @mhr3 Could you do me a favour and double check the regex, I tweaked? The idea is to more explicitly allow query parameters on the `ingest` endpoint.

Would be better to run the regex only on the path itself, not including the query.

lukasmalkmus:
> Would be better to run the regex only on the path itself, not including the query.

Why doing it easy when you can do it the hard way?! 😅 

Thanks for the hint, fixed by matching the `u.Path` of the parsed url.

mhr3 (approved):

PR #74, opened 2021-11-16
Ignore organization id when using ingest tokens

lukasmalkmus:
* Ignore the organization id when using an ingest token
* This changes some tests as the order in which the options are checked has changed. Thus, a different error is raised in some cases. But I prefer the new order since pointing out a missing access token is more obvious than pointing out a missing organizatio id.


bahlo (approved):
LGTM

PR #90, opened 2022-01-10
Support signing keys for organizations

lukasmalkmus:
Properly implement support for the upcoming shared access tokens by accepting the `keys` field.

This fixes https://github.com/axiomhq/axiom-go/runs/4755775103?check_suite_focus=true.


bahlo (approved):
lgtm

PR #92, opened 2022-01-17
Better org support

lukasmalkmus:
This PR introduces a cleaner separation of `Axiom Cloud` and `Axiom Selfhost` methods. This way, it is not possible to get confused on which methods are designated for which type of deployment.


.github/workflows/pr.yml
```diff
@@ -87,7 +87,7 @@ jobs:
     strategy:
       matrix:
         deployment:
-          - azure-1-staging
+          # - azure-1-staging
```
bahlo, line 90: Why comment out?

bahlo (approved):

.github/workflows/pr.yml
lukasmalkmus: Because parts of it are already commented out as it broke a test. In the future, we test against local Axiom anyways.

PR #104, opened 2022-02-22
Add support for '/_fields' endpoint

lukasmalkmus:
<strike>This has long been outstanding but went unnoticed: There has been a new `/_info` endpoint for a while, now. This is basically a list of dataset infos as returned by a normal call to `:id/info`. The `/_stats` call shares the same response model, apart from omitting the fields so this was never an issue. But it is desirable to have the separate call available.

I was able to refactor this slightly to compensate for the missing fields and enhanced the documentation along the way.</strike>

<strike>As explained by @mhr3, the `/_info` endpoint is deprecated and - after discussion - will be replaced by `/_fields`. The rest of the information that could be obtained through `/_info` is still available through `/_stats`. The recent commit accounts for that change but I kept the improvements to the `DatasetInfo` and `DatasetStats` types as well as the documentation improvements. Until the change is live, this PR will remain in draft status.</strike>

After more evaluation, there will be one new endpoint introduced with this PR: `/_fields`, for retrieving fields of all datasets of a deployment. However, in addition, `/_info` is deleted as it is essentially the infomration provided by `/_fields` plus some information only useful for the frontend but not actual costumers. The recent commit accounts for that change but I kept the improvements to the `DatasetInfo` and `DatasetStats` types as well as the documentation improvements. Until the change is live, this PR will remain in draft status.

Needs #106.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ endif
 
 # FUNCTIONS
 # func go-run-tool(name)
-go-run-tool = $(CGO) run -mod=mod $(shell echo $(GOTOOLS) | tr ' ' '\n' | grep -w $1)
+go-run-tool = $(CGO) run $(shell echo $(GOTOOLS) | tr ' ' '\n' | grep -w $1)
 
 .PHONY: all
 all: dep generate fmt lint test ## Run dep, generate, fmt, lint and test
--- a/axiom/apl/format.go
+++ b/axiom/apl/format.go
@@ -4,7 +4,7 @@ import (
 	"net/url"
 )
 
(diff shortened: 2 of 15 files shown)


mhr3:
We'll drop the /_info endpoint, so let's not add it here

lukasmalkmus:
This is ready for review and can be merged when CI passes on staging.

lukasmalkmus:
Updated to PR description to reflect the latest state of the discussion.

I re-ran tests and made some minor adjustments.

lukasmalkmus:
This is finally ready to merge!

bahlo (approved):
LGTM

PR #106, opened 2022-03-16
Support Go 1.18, phase out Go 1.16

lukasmalkmus:
This PR makes sure the package builds fine with the recent Go 1.18. As we just support the lst two minor releases of Go, I'll dropped support for Go 1.16. But as this only affects some toolchain related changes, it will continue to build under Go 1.16 (and possibly lower).

Needs https://github.com/golangci/golangci-lint/pull/2438 to pass CI.


kamstrup (approved):
Uh, I did not know about `t.Cleanup(f)` nice! This bump is ok by me, as long as we agree to not introduce any sweeping changes with `any` or or type params.

lukasmalkmus:
> This bump is ok by me, as long as we agree to not introduce any sweeping changes with `any` or or type params.

We won't, as `axiom-go` is supposed to support the last two minor versions of Go, which is now 1.17 and 1.18. In around half a year, when Go 1.19 comes around, we'll go that route and introduce `any` over `interface{}` and see if we can use some fancy 1.18 features. This PR is just in draft as we need an upstream fix for `golangci-lint`.

So this PR only makes sure we build fine with Go 1.18 (which was expected, tbh) and drops some of the outdated, pre 1.17 build tags.

> Uh, I did not know about `t.Cleanup(f)` nice! 

This has actually been around since Go 1.14. Pretty neat! You might also want to checkout Go 1.17's `T.Setenv` method.

PR #107, opened 2022-03-23
Support rate limit headers

lukasmalkmus:
This PR adds support for typed rate limit errors.

In addition to that, it does some cleanups on errors and their tests as well as proper types for commonly used headers and header values.

TODO:

* [ ] Integration tests


axiom/client.go
```diff
@@ -335,12 +379,67 @@ func (c *Client) do(req *http.Request, v interface{}) (*response, error) {
...
+	c.limitsMu.Unlock()
+
+	if !c.noLimiting && !limit.Reset.IsZero() && limit.Remaining == 0 && time.Now().Before(limit.Reset) {
+		// Create a fake response.
+		resp := &http.Response{
+			Status:     http.StatusText(http.StatusTooManyRequests),
+			StatusCode: http.StatusTooManyRequests,
+			Request:    req,
+			Header:     make(http.Header),
+			Body:       ioutil.NopCloser(strings.NewReader("")),
```
bahlo, line 433: Can we reuse this fake body?

axiom/client.go
```diff
@@ -246,44 +265,64 @@ func (c *Client) newRequest(ctx context.Context, method, path string, body inter
...
+		_ = httpResp.Body.Close()
+	}()
 
-	resp := &response{httpResp}
+	resp := newResponse(httpResp)
+
+	key := limitKey(resp.Limit.limitType, resp.Limit.Scope)
+	c.limitsMu.Lock()
+	c.limits[key] = resp.Limit
+	c.limitsMu.Unlock()
```
bahlo, line 320: What if you get hold of the lock after another request updated a limit lower than what we have? Do we need a check here?

README.md
```diff
@@ -48,11 +48,11 @@ When using Axiom Selfhost the URL of the deployment must be provided.
 
 ### Access Token Types
 
-The **personal token** grants access to all resources available to the user on his
-behalf.
+The **personal token** grants access to all resources available to the user on
+his behalf.
```
bahlo, line 52:
```suggestion
their behalf.
```

bahlo (approved):
LGTM, few qs

axiom/client.go
lukasmalkmus: Like `var fakeBody = ioutil.NopCloser(strings.NewReader(""))` and pass `faceBody` around?

Probably.

axiom/client.go
lukasmalkmus: You mean that request one is about to aquire the lock (and set the limit to `0`, meaning nothing remaining) while the concurrently running request two has already made it past `checkLimit()` and/or `Do()`?

In both of these cases, the limit is updated twice. However, I prefer that over dealing with additional locking as this is very unlikely to happen and does no harm. Worst case is sending a request even if a 429 has already been received.

bahlo: Well one way to do this would be wrapping line 319 in an `if resp.Limit > c.limits[key]`, right?

axiom/client.go
bahlo: Yeah probably fine without

axiom/client.go
lukasmalkmus: Like this?

```suggestion
	key := limitKey(resp.Limit.limitType, resp.Limit.Scope)
	c.limitsMu.Lock()
	if resp.Limit.Limit > c.limits[key].Limit {
		c.limits[key] = resp.Limit
	}
	c.limitsMu.Unlock()
```

I fail to see how this prevents the described problem? `resp.Limit.Limit > c.limits[key].Limit` will be `true` all the time, no?

PR #108, opened 2022-04-26
Swallow fields meta

lukasmalkmus:
This PR "swallows" the new `fieldsMeta` and `fieldsMetaMap` response objects. There is no need for them to be supported in this client library, yet. But this PR makes CI green again and documents, that the fields are there but unused.


ronoc:
Well the meta details really helped the frontend to display details, one could assume that they would also help our clients.

ronoc (approved):

PR #109, opened 2022-04-27
CI upgrades

lukasmalkmus:
* Bump most off the official actions as well as the outdated deprecated codecov one
* Bump dependencies


.github/workflows/test_examples.yaml
```diff
@@ -90,7 +90,7 @@ jobs:
     name: Examples Pass
     runs-on: ubuntu-latest
     steps:
-      - uses: LouisBrunner/checks-action@v1.1.2
+      - uses: LouisBrunner/checks-action@v1.2.0
```
a-khaledf, line 93: I think you missed a similar update to workflows/pr.yaml line 108.

lukasmalkmus: Good catch! Addressed in the latest commit.

a-khaledf (approved):
LGTM

PR #110, opened 2022-05-09
OAuth2 Auth Code + PKCE

lukasmalkmus:
This implements the OAuth2 `Authorization Code Grant` Flow (https://datatracker.ietf.org/doc/html/rfc6749#section-4.1) with `Proof Key for Code Exchange by OAuth Public Clients` (https://datatracker.ietf.org/doc/html/rfc7636#section-7.3).

Most of the OAuth2 spec wasn't really much of a concern as this is handled by `golang.org/x/oauth2`. It was mostly consulted for getting the tests working.

[RFC 7636](https://datatracker.ietf.org/doc/html/rfc7636) is what resulted in the new `axiom/auth/pkce` sub-pacakge as there is no support for this in `golang.org/x/oauth2` and implementations I considered weren't that promissing.


axiom/auth/auth.go
```diff
@@ -0,0 +1,180 @@
...
+		BaseContext: func(net.Listener) context.Context { return ctx },
+	}
+	defer srv.Close()
+
+	errCh := make(chan error)
+	go func() {
+		if serveErr := srv.Serve(lis); serveErr != nil && serveErr != http.ErrServerClosed {
+			errCh <- serveErr
+		}
+		close(errCh)
```
bahlo, line 113: This will panic if you have a servErr as it will try to close the channel here and in line 121.

axiom/auth/auth.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+
+		var buf bytes.Buffer
+		if err = tmpl.Execute(&buf, data); err != nil {
+			code := http.StatusInternalServerError
+			http.Error(w, http.StatusText(code), code)
+			return
+		}
+
+		_, _ = io.Copy(w, &buf)
```
bahlo, line 176: This can err

bahlo (commented):
LGTM otherwise

axiom/auth/auth.go
lukasmalkmus: It does close the `doneCh` in L121, not this one: https://github.com/axiomhq/axiom-go/blob/5fefd6638180af09e13497596ca3617c3b7c1731/axiom/auth/auth.go#L121

axiom/auth/auth.go
lukasmalkmus: I'm open for suggestions. I didn't want error handling to become to crazy in the callback and possible deal with another channel, here.

lukasmalkmus: Deprecated.

axiom/auth/auth_test.go
```diff
@@ -90,3 +90,43 @@ func TestLogin(t *testing.T) {
...
+			require.NoError(t, err)
+			assert.NoError(t, resp.Body.Close())
+		}()
+		return nil
+	}
+
+	token, err := auth.Login(context.Background(), srv.URL, loginFunc)
+	assert.EqualError(t, err, "oauth2: cannot fetch token: 500 Internal Server Error\nResponse: Internal Server Error\n")
+	assert.Empty(t, token)
+}
```
bahlo, line 132: Can we have a final redirect test as well?

bahlo (approved):
LGTM otherwise 🙌

PR #114, opened 2022-05-30
Fix client backoff mechanism

lukasmalkmus:
* Make sure to fix the test that is failing on `main` by not wrapping the error
* Add an additional test to make sure certain status codes are not retried
* Use the helper methods to slim down the backoff tests ;)


damaredayo (approved):
Looks good, much appreciated!

PR #120, opened 2022-07-04
feat(client): extract module info from build info for use in 'User-Agent' header

lukasmalkmus:
This extracts the module info from the build info of a binary and uses the `axiom-go` module version in the `User-Agent` header.

I tested this manually, haven't figured out how to write a test for this. Tests are still part of this module, so it won't recoginze `axiom-go` as a dependency, even if imported.

### DIFF
--- a/axiom/client.go
+++ b/axiom/client.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"os"
 	"regexp"
+	"runtime/debug"
 	"strings"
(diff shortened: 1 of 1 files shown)


bahlo (approved):

lukasmalkmus:
Tried to come up with a test but my approach didn't work. Let's not go crazy on this. Tested it manually and it worked.

PR #122, opened 2022-07-18
feat(ci): optimize

lukasmalkmus:
<strike>This PR optimizes CI by delegating caching to the `setup-go@v3` action. as per: https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs. This drops a lot of redundant caching code.</strike>

Apparently, this didn't work as expected. Leaving as is.

It also adds Go 1.19 as "experimental" to test against the upcoming of Go. The `experimental` include also improves how jobs are cancelled to reduce the amount of CI time that is used (and billed).

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -14,6 +14,7 @@ jobs:
         go:
           - 1.17
           - 1.18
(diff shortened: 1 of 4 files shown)


bahlo (approved):
LGTM

PR #125, opened 2022-07-25
fix(limit): drop faulty rate limiting logic

lukasmalkmus:
This PR drops client side rate limiting. As the Axiom service has evolved, different kinds of limits have been introduced and make implementing client-side limiting harder. As we are a service that - by nature - deals with a high volume of incoming data, we decided to not let the client deal with limiting, too, as a faulty limiting logic can prevent the client from sending data (where as the server might not apply the same limit).

### DIFF
--- a/axiom/client.go
+++ b/axiom/client.go
@@ -14,7 +14,6 @@ import (
 	"regexp"
 	"runtime/debug"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/cenkalti/backoff/v4"
@@ -70,10 +69,6 @@ type Client struct {
 	noEnv          bool
 	noLimiting     bool
 
-	// Rate limit for the client as determined by the most recent API call.
-	limits   map[string]Limit
-	limitsMu sync.Mutex
-
 	// Services for communicating with different parts of the GitHub API.
 	Datasets      *DatasetsService
 	Organizations *OrganizationsService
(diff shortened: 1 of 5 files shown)


axiom/client.go
```diff
@@ -352,9 +352,8 @@ func (c *Client) do(req *http.Request, v interface{}) (*response, error) {
 		return resp, err
 	}
 
-	key := limitKey(resp.Limit.limitType, resp.Limit.Scope)
 	c.limitsMu.Lock()
-	c.limits[key] = resp.Limit
+	c.limits[resp.Limit.limitType.String()] = resp.Limit
```
bahlo, line 356: doesn't have to be a map anymore does it?

lukasmalkmus: We still have different queries for ingest and query. `resp.Limit.limitType` is dynamic.

bahlo: Could be c.ingestLimit and c.queryLimit, no?

lukasmalkmus: Sure, but why deal with multiple fields?

bahlo: Removes the magic lookup string and is easier to grok. I won‘t die on this hill though, your decision

bahlo (dismissed):
lgtm

bahlo (approved):
lgtm

PR #127, opened 2022-08-06
Add ingest HN example, IngestChannel

bahlo:
This PR adds an example that ingests all of HN into a dataset. 
The example brought up the need of a `IngestChannel` method, which this PR also adds.


examples/ingest-hn/main.go
```diff
@@ -0,0 +1,111 @@
...
+	"log"
+	"net/http"
+	"os"
+	"strconv"
+	"sync"
+
+	"github.com/axiomhq/axiom-go/axiom"
+)
+
+const BaseURL = "https://hacker-news.firebaseio.com"
```
lukasmalkmus, line 17:
```suggestion
const baseURL = "https://hacker-news.firebaseio.com"
```

axiom/datasets.go
```diff
@@ -175,6 +176,16 @@ type IngestStatus struct {
 	WALLength uint32 `json:"walLength"`
 }
 
+func (s *IngestStatus) add(other *IngestStatus) {
```
lukasmalkmus, line 179:
```suggestion
func (s *IngestStatus) merge(other *IngestStatus) {
```

I think `add` implies values are all added up but this is not true for the `WALLength`.

axiom/datasets.go
```diff
@@ -560,6 +571,48 @@ func (s *DatasetsService) IngestEvents(ctx context.Context, id string, opts Inge
...
+			}
+		}
+	}
+	if len(eventBuf) > 0 {
+		newIngestStatus, err := s.IngestEvents(ctx, id, opts, eventBuf...)
+		if err != nil {
+			log.Fatal(fmt.Errorf("failed to ingest events: %w", err))
+		}
+
+		ingestStatus.add(newIngestStatus)
```
lukasmalkmus, line 610:
```suggestion
		ingestStatus.merge(newIngestStatus)
```

See above.

axiom/datasets.go
```diff
@@ -560,6 +571,48 @@ func (s *DatasetsService) IngestEvents(ctx context.Context, id string, opts Inge
 	return &res, nil
 }
 
+// IngestChannel ingests events from a channel into the dataset identified by
+// its id.
+// Restrictions for field names (JSON object keys) can be reviewed here:
+// https://www.axiom.co/docs/usage/field-restrictions.
+func (s *DatasetsService) IngestChannel(ctx context.Context, id string, events <-chan Event, opts IngestOptions) (*IngestStatus, error) {
```
lukasmalkmus, line 578: I think this method could use multiple cleanups:

1. It does not allow any control over the size of the buffer. I don't know if that is something users **really** need but if they are using channels it feels like they have a more complex use case that needs more control over ingestions, as well.
2. Instead of saving into a buffer we might profit from opening a connection and writing directly to the request body (using an `io.Pipe`). E.g. the whole `IngestEvents` method but ranging over the channel instead of the supplied slice of events.
3. Code seems a bit repetitive, could make the flush logic a reusable function.
4. It is logging error instead of returning them. Big no no.

examples/ingest-hn/main.go
```diff
@@ -0,0 +1,111 @@
...
+
+func getMaxItemID() (uint64, error) {
+	res, err := http.Get(BaseURL + "/v0/maxitem.json")
+	if err != nil {
+		return 0, fmt.Errorf("failed to get maxitem.json: %w", err)
+	}
+	defer res.Body.Close()
+
+	maxItemIDBytes, err := ioutil.ReadAll(res.Body)
```
lukasmalkmus, line 26:
```suggestion
	maxItemIDBytes, err := io.ReadAll(res.Body)
```

`ioutil` is deprecated.

examples/ingest-hn/main.go
```diff
@@ -0,0 +1,111 @@
...
+
+	idChan := generateIDs(maxItemID)
+	eventChan := fetchEvents(idChan)
+
+	client, err := axiom.NewClient()
+	if err != nil {
+		log.Fatal(fmt.Errorf("failed to construct Axiom client: %w", err))
+	}
+
+	_, err = client.Datasets.IngestChannel(context.TODO(), datasetName, eventChan, axiom.IngestOptions{})
```
lukasmalkmus, line 107:
```suggestion
	_, err = client.Datasets.IngestChannel(context.Background(), datasetName, eventChan, axiom.IngestOptions{})
```

Or use a proper context from `signal.NotifyContext`.

axiom/datasets.go
bahlo: Re 4.: Oops, yeah that log should not be in there.

PR #130, opened 2022-08-15
feat(query): add {min/max}Cursor to the query result status

lukasmalkmus:
This adds two missing fields.

It also improves some tests aroud the affected struct.

### DIFF
--- a/axiom/query/result.go
+++ b/axiom/query/result.go
@@ -40,6 +40,13 @@ func messageCodeFromString(s string) (mc MessageCode, err error) {
 	return mc, err
 }
 
+// MarshalJSON implements `json.Marshaler`. It is in place to marshal the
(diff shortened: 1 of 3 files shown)


axiom/query/result.go
```diff
@@ -53,7 +60,7 @@ func (mc *MessageCode) UnmarshalJSON(b []byte) (err error) {
 	return err
 }
 
-// MessagePriority represents the priority of a message associated with a query.
+// MessageCode represents the priority of a message associated with a query.
```
kamstrup, line 63: Looks like a stray search/replace?

kamstrup (approved):
Looks good, thanks @lukasmalkmus

axiom/query/result.go
lukasmalkmus: Thanks! Good catch!

PR #132, opened 2022-08-16
feat(query): add argmin, argmax and set aggregations

lukasmalkmus:
This PR adds three new aggregations to `axiom-go`:

* `argmin`
* `argmax`
* `makeset`


schehata (approved):
LGTM

PR #133, opened 2022-08-16
feat(ci): use GitHub release notes generation API for changelog

lukasmalkmus:
We use that for most of our non-Go projects as well.


schehata (approved):
looks good, not sure why the tests are failing though

PR #134, opened 2022-08-30
fix(query): parse unknown aggregations gracefully

lukasmalkmus:
Fixes an issue when running an APL query with an aggregation that has not yet been defined in `axiom-go`.

### DIFF
--- a/axiom/query/aggregation.go
+++ b/axiom/query/aggregation.go
@@ -2,7 +2,6 @@ package query
 
 import (
 	"encoding/json"
-	"fmt"
 	"strings"
 )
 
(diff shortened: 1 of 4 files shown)


bahlo (approved):
would be cool if we could still record the unknown value somewhere -- maybe the "enum" should have string values? idk, it's fine like this too, LGTM

lukasmalkmus:
> would be cool if we could still record the unknown value somewhere -- maybe the "enum" should have string values? idk, it's fine like this too, LGTM

This was my initial idea, too. But then `string("newAggOp") != OpUnknown`.

bahlo:
what if we didn't have `OpUnknown` and everything that's not typed is unknown?

PR #136, opened 2022-08-30
fix(auth/pkce): code grooming

lukasmalkmus:
Some code grooming.

### DIFF
--- a/axiom/auth/auth.go
+++ b/axiom/auth/auth.go
@@ -2,6 +2,8 @@ package auth
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"net"
@@ -9,7 +11,6 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/google/uuid"
 	"golang.org/x/oauth2"
 
 	"github.com/axiomhq/axiom-go/axiom/auth/pkce"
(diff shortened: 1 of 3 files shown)


axiom/auth/auth.go
```diff
@@ -20,8 +21,6 @@ const (
 	tokenPath   = "/oauth/token" //nolint:gosec // Sigh, this is not a hardcoded credential...
 	successPath = "/oauth/success"
 	errorPath   = "/oauth/error"
-
-	clientID = "13c885a8-f46a-4424-82d2-883cf7ccfe49"
```
lukasmalkmus, line 24: This will be moved into `cli`.

axiom/auth/auth.go
```diff
@@ -88,12 +87,13 @@ func Login(ctx context.Context, baseURL string, loginFunc LoginFunc) (string, er
 		return "", err
 	}
 
-	state, err := uuid.NewUUID()
-	if err != nil {
+	stateBytes := make([]byte, 16)
+	if _, err = rand.Read(stateBytes); err != nil {
 		return "", err
 	}
+	state := hex.EncodeToString(stateBytes)
```
bahlo, line 94: Missing some context here, why are we moving to a random hex string?

lukasmalkmus: I took some inspiration from different OAuth2 client-side implementations that utilise `state`. While my previous choice of UUID provides sufficient entropy, that approach wasn't used a lot in the wild.

Also the new approach makes it a bit more explicit where entropy is derived from as compared to the UUID approach where you actually need some insight into what an UUID is and how it works.

bahlo (approved):

PR #137, opened 2022-09-05
feat: ingest channel + example

lukasmalkmus:
This PR superseeds #127 and introduces an `IngestChannel` method that takes in a channel and ingests its content on a single request until closed.

It also includes a comprahensive example that demonstrates usage.

Closes #127.


bahlo (approved):
nice

axiom/datasets.go
```diff
@@ -331,6 +333,60 @@ func (s *DatasetsService) IngestEvents(ctx context.Context, id string, opts Inge
 	return &res, nil
 }
 
+// IngestChannel ingests events from a channel into the dataset identified by
+// its id. As it keeps a connection open until the channel is closed, it is not
+// advised to use this method for long-running ingestions.
```
kamstrup, line 338: By "not advised" we really mean that anyone not reading the docs here can easily DOS us I think? https://linear.app/axiom/issue/DB-3420/drip-feed-data-into-ingest-results-in-super-high-ingest-pressure-dos

I am a little worried exposing this as public API in our current state TBH. I think it is a nice API, but it is extremely easy to misuse unknowingly and cause us great trouble (until we fix the linked issue).

axiom/datasets.go
```diff
@@ -331,6 +333,60 @@ func (s *DatasetsService) IngestEvents(ctx context.Context, id string, opts Inge
...
+		}
+
+		var (
+			enc    = json.NewEncoder(zsw)
+			encErr error
+		)
+		for event := range events {
+			if encErr = enc.Encode(event); encErr != nil {
+				break
+			}
```
kamstrup, line 363: Do we need a check for ctx.Done() here? Reading from the events pipe can take however long and our ctx might have been cancelled...

PR #140, opened 2022-09-23
Mark BlocksDeleted as deprecated

bahlo:
@lukasmalkmus or would you rather I remove the whole `TrimResult`?


lukasmalkmus (approved):
Fine with the changes. Leave it as is for now. Before we drop this, I want to know if the server returns a 200 or 204 atm.

PR #157, opened 2022-10-20
Fix example code in README

ProZsolt:
The example in the README had some syntax errors(missing parameters, extra bracket). I fixed those.

I also mede it to a complete function

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,29 +57,36 @@ Create and use a client like this:
 
 ```go
 import (
-  "github.com/axiomhq/axiom-go/axiom"
-  "github.com/axiomhq/axiom-go/axiom/query"
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/axiomhq/axiom-go/axiom"
 )
 
-client, err := axiom.NewClient()
-if err != nil {
-  return err
-}
+func main() {
+	ctx := context.Background()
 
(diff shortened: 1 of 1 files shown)


lukasmalkmus (approved):

lukasmalkmus:
Hey @ProZsolt, so I finally had time to put some work into this: I at least made sure only tests that don't require any secrets are run from forks. But it will never be possible to satisfy the current CI from a fork for a reason:

1. We depend on two checks to pass before a PR can be merged: `CI Pass` and `Examples Pass`. This is in place to make sure a PR can be approved but does not get merged before CI is fully satisfied.
2. Those checks are created by an action that utilises `GITHUB_TOKEN`. But forks never get a token with any `write` permission. So they cannot create the check we rely on.
3. We used the extra check to aggregate the results of multiple jobs but emit only on check we can specify in the branch protection rules. This might actually be some debt left over from a previous iteration of the CI but I have to double check history to make sure.

I'll create an issue to track the progress on this and try to fix it by the end of this week, if time allows. Until then I advise organisation members to push contributions to upstream rather than a fork.

I'll leave this issue open as it is a good test bed.

lukasmalkmus (approved):

lukasmalkmus (approved):

PR #158, opened 2022-11-07
fix: correct http request retry for streams

lukasmalkmus:
This PR fixes the http retry mechanism for stream-based ingest methods by properly setting `http.Request.GetBody` as well as utilizing it. This does not apply to user-supplied streams on `client.Datasets.Ingest`, although Go will do some magic if the passed `io.Reader` is one of `bytes.Buffer`, `bytes.Reader` or `strings.Reader`.

### DIFF
--- a/axiom/datasets.go
+++ b/axiom/datasets.go
@@ -399,48 +399,38 @@ func (s *DatasetsService) IngestChannel(ctx context.Context, id string, events <
 		return nil, spanError(span, err)
 	}
 
-	getBody := func() (io.ReadCloser, error) {
-		pr, pw := io.Pipe()
+	pr, pw := io.Pipe()
 
-		zsw, wErr := zstd.NewWriter(pw)
-		if wErr != nil {
-			_ = pr.Close()
-			_ = pw.Close()
-			return nil, wErr
+	zsw, err := zstd.NewWriter(pw)
+	if err != nil {
+		_ = pr.Close()
+		_ = pw.Close()
+		return nil, err
+	}
+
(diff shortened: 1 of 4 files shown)


axiom/datasets.go
```diff
@@ -385,36 +399,48 @@ func (s *DatasetsService) IngestChannel(ctx context.Context, id string, events <
 		return nil, spanError(span, err)
 	}
 
-	pr, pw := io.Pipe()
-	go func() {
+	getBody := func() (io.ReadCloser, error) {
```
mhr3, line 402: Can't use this here, you would have lost all the previous events from the channel

lukasmalkmus: Thanks for catching this! I suspected something similar but wrote a test that passed: Turns out my test was bad :facepalm:

We can decide not to retry on that method **_or_** change how it ingests data: Instead of a steady stream of events we could decide to batch the events from the channel. That would also prevent long standing connections from slow channels. On the other hand, we have the same problem with slow readers passed to `Ingest` where we cannot batch if the data is encoded.

But implementing batching on `IngestChannel` that takes all of this into account sounds spicy. Not that it is super complicated but you want to take channel sizes, etc. into account, no?

bahlo: In axiom-rs we take chunks of 1k or whatever comes in in 1s and then send that batch. That works really well because you can retry the chunk if ingestion fails. I think it's worth it implementing that to make it "just work".

mhr3 (approved):

PR #160, opened 2022-11-09
IngestChannel batching

lukasmalkmus:
This PR adds batching support for the `IngestChannel` method. This reduces the possibility for opening connections that are open for a very long time and allows for batches to be automatically retried in case of http error.

The batching mechansim is relatively naive but should be good enough.

### DIFF
--- a/adapters/apex/apex.go
+++ b/adapters/apex/apex.go
@@ -16,10 +16,7 @@ import (
 
 var _ log.Handler = (*Handler)(nil)
 
-const (
-	batchSize    = 1024
-	sendInterval = time.Second
-)
+const defaultBatchSize = 1024
 
 // ErrMissingDatasetName is raised when a dataset name is not provided. Set it
 // manually using the SetDataset option or export `AXIOM_DATASET`.
@@ -73,7 +70,6 @@ type Handler struct {
 	ingestOptions []ingest.Option
 
 	eventCh   chan axiom.Event
-	cancel    context.CancelFunc
 	closeCh   chan struct{}
 	closeOnce sync.Once
 }
@@ -93,7 +89,7 @@ type Handler struct {
 // calling `Close()`.
 func New(options ...Option) (*Handler, error) {
 	handler := &Handler{
-		eventCh: make(chan axiom.Event, 1),
+		eventCh: make(chan axiom.Event, defaultBatchSize),
 		closeCh: make(chan struct{}),
 	}
 
@@ -120,10 +116,19 @@ func New(options ...Option) (*Handler, error) {
 		}
 	}
 
-	// Run background scheduler.
-	var ctx context.Context
-	ctx, handler.cancel = context.WithCancel(context.Background())
-	go handler.run(ctx, handler.closeCh)
+	// Run background ingest.
+	go func() {
+		defer close(handler.closeCh)
+
+		res, err := handler.client.Datasets.IngestChannel(context.Background(), handler.datasetName, handler.eventCh, handler.ingestOptions...)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to ingest events: %s\n", err)
+		} else if res.Failed > 0 {
+			// Best effort on notifying the user about the ingest failure.
+			fmt.Fprintf(os.Stderr, "event at %s failed to ingest: %s\n",
+				res.Failures[0].Timestamp, res.Failures[0].Error)
+		}
+	}()
 
 	return handler, nil
 }
@@ -133,7 +138,6 @@ func New(options ...Option) (*Handler, error) {
 func (h *Handler) Close() {
 	h.closeOnce.Do(func() {
 		close(h.eventCh)
-		h.cancel()
 		<-h.closeCh
 	})
 }
@@ -155,60 +159,4 @@ func (h *Handler) HandleLog(entry *log.Entry) error {
 	h.eventCh <- event
 
 	return nil
-}
-
-func (h *Handler) run(ctx context.Context, closeCh chan struct{}) {
-	defer close(closeCh)
-
-	t := time.NewTicker(sendInterval)
-	defer t.Stop()
-
-	events := make([]axiom.Event, 0, batchSize)
-
-	defer func() {
-		flushCtx, flushCancel := context.WithTimeout(context.Background(), 10*time.Second)
-		defer flushCancel()
-		h.ingest(flushCtx, events)
-	}()
-
-	for {
-		select {
-		case <-ctx.Done():
-			return
-		case <-t.C:
-			if len(events) == 0 {
-				continue
-			}
-		case event, ok := <-h.eventCh:
-			if !ok {
-				continue
-			}
-
-			events = append(events, event)
-
-			if len(events) < batchSize {
-				continue
-			}
-		}
-
-		h.ingest(ctx, events)
-
-		// Clear batch buffer.
-		events = make([]axiom.Event, 0, batchSize)
-	}
-}
-
-func (h *Handler) ingest(ctx context.Context, events []axiom.Event) {
-	if len(events) == 0 {
-		return
-	}
-
-	res, err := h.client.Datasets.IngestEvents(ctx, h.datasetName, events, h.ingestOptions...)
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "failed to ingest batch of %d events: %s\n", len(events), err)
-	} else if res.Failed > 0 {
-		// Best effort on notifying the user about the ingest failure.
-		fmt.Fprintf(os.Stderr, "event at %s failed to ingest: %s\n",
-			res.Failures[0].Timestamp, res.Failures[0].Error)
-	}
 }
--- a/adapters/apex/apex.go
+++ b/adapters/apex/apex.go
@@ -3,7 +3,7 @@ package apex
 import (
 	"context"
 	"errors"
-	"fmt"
+	stdlog "log"
 	"os"
 	"sync"
 	"time"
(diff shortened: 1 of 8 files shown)


axiom/datasets.go
```diff
@@ -379,70 +385,109 @@ func (s *DatasetsService) IngestEvents(ctx context.Context, id string, events []
 // Restrictions for field names (JSON object keys) can be reviewed here:
 // https://www.axiom.co/docs/usage/field-restrictions.
 //
-// As this method keeps a connection open until the channel is closed, it is not
-// advised to use this method for long-running ingestions.
+// Events are ingested in batches. A batch is either 1024 events for unbuffered
+// channels or the capacity of the channel for buffered channels. A batch is
+// sent to the server as soon as it is full or after one second or when the
+// channel is closed.
```
kamstrup, line 391: Please document under which conditions this methods returns.

axiom/datasets.go
```diff
@@ -379,70 +385,109 @@ func (s *DatasetsService) IngestEvents(ctx context.Context, id string, events []
...
+		t := time.NewTicker(defaultSendInterval)
+		defer t.Stop()
 
-	req.Header.Set("Content-Type", NDJSON.String())
-	req.Header.Set("Content-Encoding", Zstd.String())
+		for {
+			select {
+			case <-ctx.Done():
+				flushCtx, flushCancel := context.WithTimeout(context.Background(), 10*time.Second)
+				defer flushCancel()
```
kamstrup, line 447: looks dubious - as it could stack up an unlimited amount of contexts... deferred functions are called once the enclosing function return, not when it goes out of scope in the for-loop. Is that intended?

axiom/datasets.go
```diff
@@ -379,70 +385,109 @@ func (s *DatasetsService) IngestEvents(ctx context.Context, id string, events []
...
-	// Apply supplied options.
-	var opts ingest.Options
-	for _, option := range options {
-		option(&opts)
-	}
+	var (
+		errCh        = make(chan error)
+		ingestStatus ingest.Status
+	)
+	go func() {
```
kamstrup, line 408: this is a very big and complex go routine - should be broken down into some smaller chunks with clearer scoping

axiom/datasets_test.go
```diff
@@ -634,6 +633,236 @@ func TestDatasetsService_IngestChannel(t *testing.T) {
...
+			"referrer":    "-",
+			"agent":       "Debian APT-HTTP/1.3 (0.8.16~exp12ubuntu10.21)",
+		},
+	}
+
+	eventCh := make(chan Event, 2)
+	go func() {
+		for _, e := range events {
+			eventCh <- e
+			time.Sleep(time.Second + 250*time.Millisecond)
```
kamstrup, line 854: not entirely clear why this sleep is needed - pls add a comment explaining it

examples/ingesthackernews/main.go
```diff
@@ -136,7 +136,7 @@ func generateIDs(max uint64) <-chan uint64 {
 
 func fetchEvents(eventIDs <-chan uint64) <-chan axiom.Event {
 	var (
-		eventCh        = make(chan axiom.Event, maxWorkers*10)
+		eventCh        = make(chan axiom.Event, 1024)
```
kamstrup, line 139: this looks unrelated - was it intentional?

adapters/apex/apex.go
```diff
@@ -120,10 +116,19 @@ func New(options ...Option) (*Handler, error) {
...
+		defer close(handler.closeCh)
+
+		res, err := handler.client.Datasets.IngestChannel(context.Background(), handler.datasetName, handler.eventCh, handler.ingestOptions...)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to ingest events: %s\n", err)
+		} else if res.Failed > 0 {
+			// Best effort on notifying the user about the ingest failure.
+			fmt.Fprintf(os.Stderr, "event at %s failed to ingest: %s\n",
+				res.Failures[0].Timestamp, res.Failures[0].Error)
+		}
```
kamstrup, line 130: it's normally nice to have construction + running split up into 2 steps (ie this chunk would go in a Run() method where callers could decide whether or nor to use a goroutine). But I guess this is already public API that we can't change?

axiom/datasets.go
```diff
@@ -379,70 +385,109 @@ func (s *DatasetsService) IngestEvents(ctx context.Context, id string, events []
...
+		defer close(errCh)
 
-	path, err := AddOptions(s.basePath+"/"+id+"/ingest", opts)
-	if err != nil {
-		return nil, spanError(span, err)
-	}
+		// Batch is either 1024 events for unbuffered channels or the capacity
+		// of the channel for buffered channels.
+		batchSize := defaultBatchSize
+		if cap(events) != 0 {
```
bahlo, line 414: Can we make this a `<`? It makes sense for smaller channels, not so much for larger, right?

axiom/datasets.go
```diff
@@ -379,70 +385,109 @@ func (s *DatasetsService) IngestEvents(ctx context.Context, id string, events []
...
-	zsw, err := zstd.NewWriter(pw)
-	if err != nil {
-		_ = pr.Close()
-		_ = pw.Close()
-		return nil, err
-	}
+			// Reset batch buffer after attempted ingest, regardless of the
+			// outcome.
+			defer func() {
+				batch = make([]Event, 0, batchSize)
```
bahlo, line 427: We could trim the slice instead (or do we have to create a copy?)

```suggestion
				batch = batch[:0]
```

axiom/datasets.go
bahlo: +1, maybe wrap it in a func() {} to make use of the defer?

bahlo (changes requested):
some nitpicks, but lgtm overall

axiom/datasets.go
lukasmalkmus: > Is that intended?

Yes, it is. Each of the `defer` statements is followed by a `return`. A previous design had a single `defer` function which was responsible for flushing the remaining events if the function returns. But there are different requirements for both cases: When returning because the context is marked as done, I want to return the context error. When returning because the channel is closed, I want to return no error. Thus the two different occasions.

lukasmalkmus: > +1, maybe wrap it in a func() {} to make use of the defer?

Correct me if I'm wrong but the `defer` is indeed used and does not stack up?!

examples/ingesthackernews/main.go
lukasmalkmus: Intentional to match the default batch size.

adapters/apex/apex.go
lukasmalkmus: You think so? I figure people want a seamless setup when setting up their logging: https://github.com/axiomhq/axiom-go/blob/main/examples/apex/main.go#L14-L17.

> _[...]_ where callers could decide whether or nor to use a goroutine _[...]_

Is there really a decision to be made? They need this running concurrently, no?

> But I guess this is already public API that we can't change?

We have yet to release a stable version. We can change everything.

axiom/datasets.go
lukasmalkmus: But if the slice is extended again, the original data reappears, no? Which shouldn't be a problem as we don't extend it.

axiom/datasets.go
lukasmalkmus: Tbh., now I wonder if we should actually deal with the extra context? Maybe we should obey `ctx` 🤔

lukasmalkmus: I refactored the whole method and this shouldn't be a problem anymore. Please re-review.

axiom/datasets.go
lukasmalkmus: I refactored the whole method and this shouldn't be a problem anymore. Please re-review.

I noticed that I don't need a go-routine as the method is blocking, anyways. The idea with the go-routine came from another design I had in mind where a go-routine would drain the producer channel into a batch and the parent go-routine would swap the full batch with an empty one so it can ingest without blocking the producer channel being read.

axiom/datasets.go
```diff
@@ -379,70 +385,80 @@ func (s *DatasetsService) IngestEvents(ctx context.Context, id string, events []
...
-			// one.
-			encErr = closeErr
-		}
-		_ = pw.CloseWithError(encErr)
-	}()
-
-	req, err := s.client.NewRequest(ctx, http.MethodPost, path, pr)
-	if err != nil {
-		return nil, spanError(span, err)
+		return nil
```
kamstrup, line 435: Reads easier with early return to avoid unnecessary indentation
```
if len(batch) == 0 {
    return nil
}

do the ingest
```

kamstrup (commented):
New IngestChannel func is a big improvement, nice stuff Lukas! <3

adapters/apex/apex.go
lukasmalkmus: Makes sense. I'll do this in a follow up.

adapters/apex/apex.go
```diff
@@ -120,10 +116,21 @@ func New(options ...Option) (*Handler, error) {
...
 
-	// Run background scheduler.
-	var ctx context.Context
-	ctx, handler.cancel = context.WithCancel(context.Background())
-	go handler.run(ctx, handler.closeCh)
+	// Run background ingest.
+	go func() {
+		defer close(handler.closeCh)
+
+		logger := stdlog.New(os.Stderr, "[AXIOM|APEX]", 0)
```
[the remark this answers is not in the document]
lukasmalkmus: Not yet. But might be a good idea 👍 Will do in a follow up.

axiom/datasets.go
lukasmalkmus: What about allowing a maximum of `1024 * 10`?

bahlo: LOL my comment doesn't make sense at all, sorry about that. Yeah basically if `cap(events) != 0 && cap(events) < batchSize`. Also please do 1000, that's what we use in lots of other contexts

bahlo (approved):
lgtm

PR #172, opened 2022-11-18
feat: decouple 'query' and 'querylegacy' package

lukasmalkmus:
Decouple `query` from the deprecated `querylegacy` package.


bahlo (approved):
we need to make sure the query method change is communicated visibly in the release, see [axiom-node](https://github.com/axiomhq/axiom-node/releases/tag/v0.10.0)

PR #173, opened 2022-11-18
More GoDoc improvements

lukasmalkmus:
Makes use of enhanced Go 1.19 formatting to present nicer GoDoc's.

Focus in this PR was on linking concrete types in GoDoc comments and creating better links, lists and code blocks.

Some comments have been cleaned up to match similar comments in their wording and/or layout.


bahlo (approved):
nice

PR #179, opened 2022-11-24
Top-level aliases for ingest and query methods

lukasmalkmus:
This PR adds aliases for ingest and query methods on the `Client`. It is just an alias and uses the implementation on the `DatasetsService` but it keeps the full GoDoc comment for ease-of-use with autocomplete functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,14 +79,14 @@ func main() {
         log.Fatalln(err)
     }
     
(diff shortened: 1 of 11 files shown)


bahlo (changes requested):
please move them to top-level instead of aliasing ❤️

bahlo (approved):

PR #184, opened 2022-12-14
feat: cloud.axiom.co -> {api,app}.axiom.co

lukasmalkmus:
Also drops the `auth` package which will be moved to CLI.

Needs #185.

Closes #DX-422.


lukasmalkmus:
While some tests are failing, this is because we depend on Axiom CLI which does not yet understand `api.axiom.co`. So after this is reviewed, we have to merge this by bypassing the example tests. Other integration tests work just fine.

mhr3 (approved):

PR #186, opened 2022-12-15
feat(query): cursor support

lukasmalkmus:


lukasmalkmus:
This PR is only adding cursor support (because it is indeed useful), but does not yet provide a higher level streaming function as this only makes sense when we have the new results format.

bahlo (commented):
lgtm, does this work on prod already?

bahlo (approved):

PR #190, opened 2023-01-24
fix: ignore field added to legacy query response

lukasmalkmus:
In preparation for the new tabular result format, the "legacy" query response got a new field denoting the response as "legacy".

This "unbreaks" the scheduled CI job.


bahlo (approved):
ok cool

PR #201, opened 2023-03-20
feat(datasets): support event labels

lukasmalkmus:
Support for event labels that are applied on every event that is part of the ingest request.


mhr3 (approved):
LGTM

PR #208, opened 2023-04-12
fix(test): ignore 'blocksExamined' in query result status

lukasmalkmus:
This assertion was faulty and in the future `blocksExamined` will be dropped from the API, anyways.


axiom/datasets_integration_test.go
mschoch, line 1: The legacy query on line 224 makes a similar assertion that should probably also be removed.

mschoch (approved):
One suggestion, otherwise looks good

axiom/datasets_integration_test.go
lukasmalkmus: Good catch, appreciated. That's what reviews are for 😄

PR #209, opened 2023-04-13
fix(apl/test): cursors

lukasmalkmus:
Fixes integration tests and and GoDoc on cursors in APL.

### DIFF
--- a/axiom/datasets_integration_test.go
+++ b/axiom/datasets_integration_test.go
@@ -327,37 +327,57 @@ func (s *DatasetsTestSuite) TestCursor() {
 	// retrieved from the query status MinCursor or MaxCursor fields, depending
 	// on the queries sort order.
 	midRowID := queryResult.Matches[1].RowID
-	midTime := queryResult.Matches[1].Time
 
-	// Query events with a cursor in descending order.
+	// Query events with a cursor in descending order...
 	apl = fmt.Sprintf("['%s'] | sort by _time desc", s.dataset.ID)
 	queryResult, err = s.client.Datasets.Query(s.ctx, apl,
-		// HINT(lukasmalkmus): Start time is inclusive when sorting newest
-		// first.
-		query.SetStartTime(midTime.Add(-time.Nanosecond)),
+		query.SetStartTime(startTime),
+		query.SetEndTime(endTime),
+		query.SetCursor(midRowID, false),
+	)
+	s.Require().NoError(err)
+
+	if s.Len(queryResult.Matches, 1) {
+		s.Equal("bar", queryResult.Matches[0].Data["foo"])
+	}
+
+	// ...again, but with the cursor inclusive.
+	queryResult, err = s.client.Datasets.Query(s.ctx, apl,
+		query.SetStartTime(startTime),
 		query.SetEndTime(endTime),
 		query.SetCursor(midRowID, true),
 	)
 	s.Require().NoError(err)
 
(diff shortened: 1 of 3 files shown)


axiom/datasets_integration_test.go
```diff
@@ -327,37 +327,57 @@ func (s *DatasetsTestSuite) TestCursor() {
...
 
-	// Query events with a cursor in descending order.
+	// Query events with a cursor in descending order...
 	apl = fmt.Sprintf("['%s'] | sort by _time desc", s.dataset.ID)
 	queryResult, err = s.client.Datasets.Query(s.ctx, apl,
-		// HINT(lukasmalkmus): Start time is inclusive when sorting newest
-		// first.
-		query.SetStartTime(midTime.Add(-time.Nanosecond)),
+		query.SetStartTime(startTime),
+		query.SetEndTime(endTime),
```
kamstrup, line 335: Just to understand what's going on here: the rowID here is actually the full encoded cursor with timestamp and "real" rowID? So server side we end up overwriting the time boundaries from it?

In normal "low level usage" I think the start time should also be set from the row that we grab the rowID from -- so even though it was clunky before, I think it was also more correct -- but as I suggest above, I think the rowID in this test is more than just rowID?

axiom/datasets_integration_test.go
```diff
@@ -327,37 +327,57 @@ func (s *DatasetsTestSuite) TestCursor() {
...
+		query.SetCursor(midRowID, false),
+	)
+	s.Require().NoError(err)
+
+	if s.Len(queryResult.Matches, 1) {
+		s.Equal("bar", queryResult.Matches[0].Data["foo"])
```
kamstrup, line 341: This is all a bit hard to follow, but I think it is correct as you have it. Maybe add a comment ala `// "buz", and "baz" skipped by cursor, so only oldest item left`

axiom/query/options.go
```diff
@@ -20,22 +18,18 @@ type Options struct {
 // An Option applies an optional parameter to a query.
 type Option func(*Options)
 
-// SetStartTime specifies the query intervals start time. If sorting with newest
-// first the value is inclusive, otherwise exclusive.
+// SetStartTime specifies the query intervals start time.
```
kamstrup, line 21:
```suggestion
// SetStartTime specifies the start time of the query interval.
```

axiom/query/options.go
```diff
@@ -20,22 +18,18 @@ type Options struct {
...
 func SetStartTime(startTime time.Time) Option {
 	return func(o *Options) { o.StartTime = startTime }
 }
 
-// SetEndTime specifies the query intervals end time. If sorting with newest
-// first the value is exclusive, otherwise inclusive.
+// SetEndTime specifies the query intervals end time.
```
kamstrup, line 26: Rephrase like my previous suggestion or fix "intervals" -> "interval's" ;-)

kamstrup (approved):
Nice w a simplification - a bit easier to read now. Mostly just some clarifying questions

axiom/datasets_integration_test.go
lukasmalkmus: > Just to understand what's going on here: the rowID here is actually the full encoded cursor with timestamp and "real" rowID? So server side we end up overwriting the time boundaries from it?

Yes, e.g. `0cs1n52vmf2f4-06580c6bde001730-0001`. Just taken from the previous query result.

PR #210, opened 2023-04-17
Add support for APL variables

bahlo:


ted-gould (dismissed):

ted-gould:
Just gonna note that I don't think these are in production yet. So we probably don't want this to land until it is fully there.

axiom/query/options.go
```diff
@@ -33,3 +35,19 @@ func SetEndTime(endTime time.Time) Option {
 func SetCursor(cursor string, include bool) Option {
 	return func(o *Options) { o.Cursor = cursor; o.IncludeCursor = include }
 }
+
+// SetVariables specifies variables which can be referenced by the APL query.
+func SetVariables(variables map[string]any) Option {
+	return func(o *Options) { o.Variables = variables }
+}
+
+// SetVariable sets a single variable which is can be referenced by the APL
```
lukasmalkmus, line 44:
```suggestion
// SetVariable sets a single variable which can be referenced by the APL
```

lukasmalkmus (approved):
LGTM, please review suggestion.

PR #217, opened 2023-04-25
chore: tune http client

lukasmalkmus:


axiom/client.go
```diff
@@ -57,6 +57,7 @@ type service struct {
 func DefaultHTTPClient() *http.Client {
 	return &http.Client{
 		Transport: DefaultHTTPTransport(),
+		Timeout:   time.Hour,
```
mhr3, line 60: Drop pls, this can be controlled better with request contexts plus a timeout this long doesn't really help anyone.

mhr3 (approved):

PR #240, opened 2023-07-18
feat(adapters): slog

lukasmalkmus:
An adapter for `golang.org/x/exp/slog` that will become `log/slog` in Go 1.21. We will transition to `log/slog` once Go 1.22 is out so we can remain compatible with the two latest minor releases of Go.

This adapters implementation is close to the `apex` and `logrus` ones. I could have utilised the `slog.JSONHandler` that would just produce JSON that the adapter needs to ingest. It would have probably yielded slightly more performance but with more complex synchronisation primitives and less elaborated batching behaviour (or investing more time in implementing that).

Let's keep it simple, for now.


adapters/slog/README.md
```diff
@@ -0,0 +1,50 @@
...
+        axiom.SetPersonalTokenConfig("AXIOM_TOKEN", "AXIOM_ORG_ID"),
+    ),
+)
+```
+
+### ❗ Important ❗
+
+The adapter uses a buffer to batch events before sending them to Axiom. This
+buffer must be flushed explicitly by calling [Close](https://pkg.go.dev/github.com/axiomhq/axiom-go/adapters/slog#Handler.Close).
+Checkout the [example](../../examples/slog/main.go).
```
bahlo, line 50: Can we add a code example with a `func main()` that sets up the adapter, logs using slog and then closes the adapter on exit?

adapters/slog/slog.go
```diff
@@ -0,0 +1,256 @@
...
+func SetClient(client *axiom.Client) Option {
+	return func(h *Handler) error {
+		h.client = client
+		return nil
+	}
+}
+
+// SetClientOptions specifies the Axiom client options to pass to
+// [axiom.NewClient] which is only called if no [axiom.Client] was specified by
+// the [SetClient] option.
```
bahlo, line 39: Would it be better DX if we added `SetToken` directly without needing to know the concept of a client or an org id?

adapters/slog/slog.go
```diff
@@ -0,0 +1,256 @@
...
+		minLevel = h.level.Level()
+	}
+	return level >= minLevel
+}
+
+// Handle implements [slog.Handler].
+func (h *Handler) Handle(_ context.Context, r slog.Record) error {
+	event := axiom.Event{}
+
+	// Set record attributes first, handler attributes second.
```
bahlo, line 183: Is that the correct order? I would expect record attributes to override handler attributes.

adapters/slog/slog.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	// Set timestamp, level and actual message. The zero time is ignored.
+	if !r.Time.IsZero() {
+		event[ingest.TimestampField] = r.Time.Format(time.RFC3339Nano)
+	}
+	event[slog.LevelKey] = r.Level.String()
+	event[slog.MessageKey] = r.Message
+
+	h.eventCh <- event
```
bahlo, line 204: If the channel is closed this will panic, can we not panic?

adapters/slog/README.md
lukasmalkmus: Like the one that is part of this PR? 🙃

adapters/slog/slog.go
lukasmalkmus: We could but I guess that is out of scope for this PR. The options are something that is duplicated code across all adapters. Maybe generic would come in handy here.

adapters/slog/slog.go
lukasmalkmus: This isn't really specified AFAIK. Do you have a strong opinion on this?

adapters/slog/slog.go
lukasmalkmus: Good call, changed this for all affected adapters.

adapters/slog/slog.go
bahlo: Yes, record > handler

adapters/slog/README.md
bahlo: Yeah, but not in the README

adapters/slog/slog.go
bahlo: Imo duplication is fine in this case to make it more usable

adapters/slog/README.md
lukasmalkmus: We have them in the GoDoc. I understand your idea but that means we need to maintain three examples for each adapter (GoDo, examples folder and adapter README).

bahlo: aight

bahlo (approved):
good

PR #242, opened 2023-07-28
chore(options): don't panic when 'nil' is passed as an option

lukasmalkmus:
Just a simple quality of life improvent that prevents users from cauing a panic when accidentally passing `nil` as a configuration option.


kevinehosford (approved):
🤷‍♂️

PR #251, opened 2023-08-17
update log slog version in go 121

thecraftman:
- Update log/slog adapter to support go 1.21

The request came in as a community discussion: https://discord.com/channels/1065957163161370664/1065957163933114411/1141490534259621998


lukasmalkmus:
@thecraftman We cannot update to Go 1.21 (yet) as we need to maintain compatibility with older versions. I addresses the problem here: #252.

PR #252, opened 2023-08-17
feat(adapters): log/slog

lukasmalkmus:
This PR enables support for the standard libarary `log/slog` package (and retains support for the non-std package by renaming to `slogx` until we upgrade to Go 1.22 (and thus only retain support down to Go 1.21).


bahlo (approved):
good

PR #256, opened 2023-09-01
Code grooming

lukasmalkmus:
Two small code grooming tasks.


adapters/apex/apex_test.go
```diff
@@ -101,13 +101,13 @@ func TestHandler_FlushFullBatch(t *testing.T) {
...
 	// Let the server process.
-	time.Sleep(250 * time.Millisecond)
+	time.Sleep(time.Millisecond * 250)
 
 	// Should have a full batch right away.
 	assert.EqualValues(t, 1000, atomic.LoadUint64(&lines))
 
 	// Wait for timer based handler flush.
-	time.Sleep(1250 * time.Millisecond)
+	time.Sleep(time.Millisecond * 1250)
```
ImLunaHey, line 110: Honestly just curious. Is there any reason `1_250` isn't used? 🤔

lukasmalkmus: The modernised number literals were introduced with Go 1.13. I guess I worked on this SDK before that released and never picked up the newer syntax.

Tbh. I feel like this could even be rather confusing for new contributors who aren't familiar with Go code. Go code should be simple and I haven't seen that syntax used a lot in the wild. I agree it would be more helpful for longer numbers, e.g. `1_000_000` is easier to read than `1000000`... But I don't think this applies to literals with less than five characters and different digits. And when dealing with duration in Go, one usually uses `time.Duration` and the constants provided by that package and I don't think I have ever seen `time.Millisecond` multiplied by an integer literal with more than 4 or 5 digits.

ImLunaHey: Gotcha. I can't speak to go but for typescript this is very common nowadays. I believe ts has had it a fair bit longer so that'd explain the difference in usage.

Personally I have dyscalculia and find even with 4 digits it helps a **LOT** having the underscore. I can see how it could be confusing to newer devs though.

lukasmalkmus: I don't have a very string opinion on this, tho. If you're saying using underscores improves accessibility, we can go for it.

lukasmalkmus: @ImLunaHey Do you have a strong opinion on `time.Millisecond * 1_250` vs. `time.Second + time.Millisecond * 250`?

ImLunaHey: How does math ordering work in go? 😓 

I tend to err on the side of caution with all languages when it comes to math and always add parentheses even if they're not "needed".

In terms of which one of those I prefer I like the first one more. Completely personal preference though.

lukasmalkmus: > How does math ordering work in go? 😓

`*` before `+` :)

See "**Operator precedence**": https://go.dev/ref/spec#Operators.

schehata (approved):

PR #258, opened 2023-09-05
chore: readme grooming

lukasmalkmus:
@bahlo Something for you. I like the overall cleaner look instead of using emojis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,11 +58,12 @@ client, err := axiom.NewClient(
 )
 ```
 
-❗ _When only performing **ingest** or **query** operations, we recommend using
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -58,11 +58,12 @@ client, err := axiom.NewClient(
 )
 ```
 
-❗ _When only performing **ingest** or **query** operations, we recommend using
+> [!IMPORTANT]
```
bahlo, line 61:
```suggestion
> [!NOTE]
```

This should be note, it's a recommendation, nothing more.

bahlo (approved):
nice ✨

PR #259, opened 2023-09-12
feat: expose axiom trace id

lukasmalkmus:
Exposes the trace id for every error and in the ingest and query result types.


axiom/datasets.go
```diff
@@ -499,6 +507,10 @@ func (s *DatasetsService) IngestEvents(ctx context.Context, id string, events []
 // The method returns without an error if the channel is closed and the buffered
 // events are successfully sent to the server.
 //
+// The returned ingest status does not contain a trace ID as the underlying
```
tsenart, line 510: I guess it's not worth aggregating all the trace IDs and exposing a slice?

axiom/query/result.go
```diff
@@ -140,6 +140,9 @@ type Result struct {
 	// GroupBy is a list of field names to group the query result by. Only valid
 	// when at least one aggregation is specified.
 	GroupBy []string `json:"-"`
+	// TraceID is the ID of the trace that was generated by the server for this
+	// results query request.
+	TraceID string `json:"-"`
```
tsenart, line 145: On my side I will need this marshaled out, so I'll just use a custom result type.

axiom/response.go
```diff
@@ -6,6 +6,8 @@ import "net/http"
 type Response struct {
 	*http.Response
 
+	TraceID string
```
tsenart, line 9: Since we embed the *http.Response, and we can get the TraceID from the header, I think it would be nicer to just add a method to Response like this:

```go
func (r *Response) TraceID() string {
    return r.Header.Get(headerTraceID)
}
```

tsenart (approved):

axiom/datasets.go
lukasmalkmus: I thought about that, as well. It feels a bit.. odd? Don't you think? I guess people might want to use the TraceID in their logging or own instrumenting. From a user perspective, dealing with a `string` is way easier than dealing with a slice (e.g. it can be `nil`). IngestChannel is a nice helper but not exactly everybody's first choice. You still get the TraceID from lower-level methods. Given that users probably want to hand it to use for investigating performance problems, unexpected behaviour, etc. they most likely need a reproducer first, that should ideally be a single request. IngestChannel still isn't a blackbox: It is instrumented via OpenTelemetry.

But Talking about OpenTelemetry: We could add the `axiom_trace_id` to every span? If people ever see something odd, they have the TraceID at hand?

axiom/datasets.go
```diff
@@ -363,10 +363,14 @@ func (s *DatasetsService) Ingest(ctx context.Context, id string, r io.Reader, ty
...
-	var res ingest.Status
-	if _, err = s.client.Do(req, &res); err != nil {
+	var (
+		res  ingest.Status
+		resp *Response
+	)
+	if resp, err = s.client.Do(req, &res); err != nil {
 		return nil, spanError(span, err)
 	}
+	res.TraceID = resp.TraceID()
```
tsenart, line 373: Do you think that there might be other things we'll want to extract from the Response down the line? Would it be too much to embed `*Response` in all these `Status` structs so that one can then just call `TraceID()`?

tsenart (approved):

axiom/datasets.go
lukasmalkmus: `Response` itself embeds `*http.Response` so that would be exported to our users, as well. I don't like that idea too much, tbh as I prefer the very cleaned-up, self-explaining structs we export.

In the future we might want to clean that up but lets re-visit that in a follow-up PR.

PR #274, opened 2023-12-05
Added an adapter for zerolog

Rambatino:
This PR implements the adapter for Zerolog: https://github.com/rs/zerolog

I tried using it's hook, similar to the logrus implementation, however the properties are not accessible for performance reasons, see: https://github.com/rs/zerolog/issues/93

I opted to implement it in a similar way to the DataDog zerolog adapter: https://github.com/archdx/zerolog-sentry/blob/master/writer.go

It generally behaves quite similar, in short we create a multiwriter, such that it writes to the zerolog adapater as well as stderr (or wherever) and we parse that json into the axiom.Event struct and pass that to the ingest client. 

It does mean the Close() func must have an error (so it differs slightly to, say, the logrus implementation)


Rambatino:
@lukasmalkmus sorry for the delay! Have been trying to find time to do this.

So I've opted to use channels for communication around the piped Write() as opposed to using mutex locks and having a package wide buffer as was the case for zap.go. But this now has the same functionality as IngestEvents whereby there is a 1000 record buffer before sending as zst encoded NDJSON.

zibens:
Pls merge, this is so helpful.

adapters/zerolog/zerolog.go
```diff
@@ -0,0 +1,286 @@
...
+
+			lvl, err := zerolog.ParseLevel(lvlStr)
+			if err != nil {
+				logger.Printf("failed to parse level: %s\n", err)
+				continue
+			}
+
+			if _, enabled := w.levels[lvl]; !enabled {
+				continue
+			}
```
lukasmalkmus, line 260: I wonder: Is there no way to let `zerolog` control which levels get send to our adapter?

adapters/zerolog/zerolog.go
```diff
@@ -0,0 +1,286 @@
...
+			timeStr, err := jsonparser.GetUnsafeString(data, zerolog.TimestampFieldName)
+			if err != nil {
+				// if Timestamp not set on the logger, default to time.Now()
+				timeStr = zerolog.TimestampFunc().Format(zerolog.TimeFieldFormat)
+			}
+
+			// ignore any errors here
+			data, _ = jsonparser.Set(data, []byte("\""+timeStr+"\""), ingest.TimestampField)
+			data, _ = jsonparser.Set(data, []byte("\""+lvlStr+"\""), zerolog.LevelFieldName)
+			data, _ = jsonparser.Set(data, loggerName, "logger")
```
lukasmalkmus, line 271: I wonder what's going on here. Isn't the timestamp guaranteed to be there when logged from `zerolog`?

Also there is no need to rewrite the timestamp field, no? We can just tell the Axiom client the name of the timestamp field as well as the timestamp format: https://github.com/axiomhq/axiom-go/blob/main/axiom/ingest/options.go#L30-L41.

lukasmalkmus (commented):
Left you some comments. Don't worry, it's not a lot of crazy stuff I'm asking.

I noticed this is still missing integration testing. There isn't a lot of useful info in the documentation so I definitely have to add that!

Basically, all integration tests are in separate `_integration_test.go` files and behind the `integration` build tag (run `make test-integration` or see the `Makefile` so check how `go test` is invoked). Config is picked up from the environment. So basically you want to point Axiom CLI to the correct org a run `eval $(axiom config export -f)` to export credentials to your env. Then you have to export `AXIOM_DATASET` and point it to a valid dataset you want to use for testing. Then you just have to invoke `go test -tags=integration`. In CI, all of that is done automatically.

examples/zerolog/main.go
```diff
@@ -0,0 +1,30 @@
...
+
+func main() {
+	// Export "AXIOM_DATASET" in addition to the required environment variables.
+
+	writer, err := adapter.New()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	l.Logger = zerolog.New(io.MultiWriter(writer, os.Stderr)).With().Timestamp().Logger()
```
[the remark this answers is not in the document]
Rambatino: So it's established in the sense I saw the datadog adapter do it. But that's about it, I tried to find another way similar to the other logging libraries, but didn't find a solve. That's not to say there isn't one tho...

adapters/zerolog/zerolog.go
```diff
@@ -0,0 +1,286 @@
...
+type optionFunc func(*WriteCloser)
+
+func (fn optionFunc) apply(c *WriteCloser) { fn(c) }
+
+// WithClient configures a custom axiom client.
+func WithClient(client *axiom.Client) writerOption {
+	return optionFunc(func(cfg *WriteCloser) {
+		cfg.client = client
+	})
+}
```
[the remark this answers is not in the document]
Rambatino: Can conform for sure!

adapters/zerolog/zerolog.go
```diff
@@ -0,0 +1,286 @@
...
+		if err != nil {
+			return fmt.Errorf("failed to ingest events: %w", err)
+		}
+		if res.Failed > 0 {
+			return fmt.Errorf("failed to ingest events: %w", err)
+		}
+		if res.Failed > 0 {
+			logger.Printf("event(s) [%v] at %s failed to ingest: %s\n", res.Failed, res.Failures[0].Timestamp, res.Failures[0].Error)
+			return nil
+		}
```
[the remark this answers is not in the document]
Rambatino: ahh yes bit of an oversight

adapters/zerolog/README.md
```diff
@@ -0,0 +1,57 @@
...
+    "github.com/axiomhq/axiom-go/axiom"
+    adapter "github.com/axiomhq/axiom-go/adapters/zerolog"
+)
+
+// ...
+client, err := axiom.NewClient()
+if err != nil {
+    log.Fatalln(err)
+}
+writer, err := adapter.New(WithDatasetName("logs"), WithAxiomClient(client))
```
lukasmalkmus, line 46: I think we renamed those options to `Set...`, no? Also make sure to add the package name before each option. Ideally, make sure this more or less matches your example code from `examples/` as that explicitly has to import the package :)

adapters/zerolog/zerolog.go
```diff
@@ -0,0 +1,255 @@
...
+// "Set".
+//
+// An API token with "ingest" permission is sufficient enough.
+//
+// A Writer needs to be closed properly to make sure all logs are sent by calling
+// [Writer.Close].
+func New(opts ...Option) (*Writer, error) {
+	w := &Writer{
+		levels: make(map[zerolog.Level]struct{}),
+		// @todo: test if this is set and not passed does axiom error
```
lukasmalkmus, line 129: Has this been resolved?

adapters/zerolog/README.md
lukasmalkmus: I also think you should drop the `NewClient(...)` and `SetClient(...)` code from the example. The adapter will auto-create a client under the hood if no client is set. Makes the example much lighter.

lukasmalkmus (approved):
Some nitpicks, otherwise LGTM.

adapters/zerolog/README.md
Rambatino: 👍

adapters/zerolog/zerolog.go
Rambatino: it has! 👍 will remove

zibens:
you guys are kings.  ty.  just added to my codebase.

lukasmalkmus:
Feel free to provide us with feedback :)

PR #285, opened 2024-02-15
chore: testing improvements

lukasmalkmus:
* Dropped the `integration` build tag. Integration tests are now run when the `AXIOM_` env vars are present.
* Found some gaps in the test coverage that I addressed
* Added test shuffling
* Had to perform some minor refactoring to deal with the removal of the `integration` build tag


axiom/otel/trace_integration_test.go
```diff
@@ -9,43 +7,83 @@ import (
...
+	"github.com/axiomhq/axiom-go/internal/config"
+	"github.com/axiomhq/axiom-go/internal/test/testhelper"
 )
 
 func TestTracingIntegration(t *testing.T) {
-	ctx := context.Background()
+	cfg := config.Default()
+	if err := cfg.IncorporateEnvironment(); err != nil {
+		t.Fatal(err)
+	}
```
peterbourgon, line 25: Would it make sense to `require.NoError(t, cfg.IncorporateEnvironment())` here (and in similar places elsewhere)?

axiom/otel/trace_integration_test.go
```diff
@@ -9,43 +7,83 @@ import (
...
+
+	if cfg.Token() == "" || cfg.OrganizationID() == "" || cfg.BaseURL() == nil {
```
peterbourgon, line 27: BaseURL seems to default to the proper Axiom API URL, so would it ever be non-nil?

axiom/otel/trace_integration_test.go
```diff
@@ -9,43 +7,83 @@ import (
...
+		t.Skip("missing required environment variables to run integration tests")
```
peterbourgon, line 28: Any way to enumerate the expected environment variables in this Skip message?

(Same comment for other t.Skip calls elsewhere.)

peterbourgon (commented):
Seems OK, at a high level.

If you have `AXIOM_` env vars set which grant read/write access to an e.g. prod environment, it seems like running `go test ./...` will now create and ingest-to and read-from and (hopefully!) delete data in that prod environment?

My concern is basically that these test (might?) be looking at the same env vars that are used by the normal Axiom CLI, rather than test-specific env vars. Those env vars may be set "by default" by a user that might end up running these tests, and in that case wouldn't really represent the true user "intent" -- they probably wouldn't want to run integration tests against those e.g. prod environments. If that's correct, would it make sense for the tests to use different env vars than the defaults?

lukasmalkmus:
> If you have `AXIOM_` env vars set which grant read/write access to an e.g. prod environment, it seems like running `go test ./...` will now create and ingest-to and read-from and (hopefully!) delete data in that prod environment?

Yes.

> My concern is basically that these test (might?) be looking at the same env vars that are used by the normal Axiom CLI, rather than test-specific env vars.

That was a nice effect of the build tag approach, integration tests had to be run explicitly. I chose the same env vars as CLI as this makes setting them up for testing is super easy: `eval $(axiom config export -f)`.

> Those env vars may be set "by default" by a user that might end up running these tests, and in that case wouldn't really represent the true user "intent" -- they probably wouldn't want to run integration tests against those e.g. prod environments. If that's correct, would it make sense for the tests to use different env vars than the defaults?

I thought about adding something like `AXIOM_INTEGRATION_TESTS` for enabling them explicitly but that felt counter-intuitive towards there goal of this PR. But I really would like keep the `AXIOM_` prefixed env vars.

Any other ideas? 🤔

peterbourgon:
It's tricky. Normally I would suggest using entirely different env vars to specify config for integration tests, versus env vars that specify config for normal execution of the program. But in this case that's difficult, because so much of the actual/real code is reading config from env vars directly, rather than having that config explicitly provided during construction.

In this case, I think I would definitely gate the integration tests on an e.g. `AXIOM_INTEGRATION_TESTS=true` env var, as you suggested. The `t.Skip` message should note this, as well as any other env vars that need to be set, to make the tests work as intended. For example,

```go
t.Skip("set AXIOM_INTEGRATION_TESTS=true AXIOM_URL=<URL> AXIOM_TOKEN=<TOKEN> ... to run this test")
```

lukasmalkmus:
> It's tricky. Normally I would suggest using entirely different env vars to specify config for integration tests, versus env vars that specify config for normal execution of the program. But in this case that's difficult, because so much of the actual/real code is reading config from env vars directly, rather than having that config explicitly provided during construction.

This is opt-out. The integration tests actually do read the env vars and set up the client explicitly rather than using its "auto-env" config mechanism: https://github.com/axiomhq/axiom-go/blob/lukasmalkmus/testing-improvements/axiom/client_integration_test.go#L28-L37.

> In this case, I think I would definitely gate the integration tests on an e.g. `AXIOM_INTEGRATION_TESTS=true` env var, as you suggested. The `t.Skip` message should note this, as well as any other env vars that need to be set, to make the tests work as intended. For example,
> 
> ```go
> t.Skip("set AXIOM_INTEGRATION_TESTS=true AXIOM_URL=<URL> AXIOM_TOKEN=<TOKEN> ... to run this test")
> ```

+1

peterbourgon:
> This is opt-out. 

OK, then first order of business would be to make it explicitly opt-in. And the presence of e.g. `AXIOM_TOKEN` doesn't necessarily mean that the caller of `go test ./...` intended to use that token to run integration tests! 😅

PR #294, opened 2024-04-02
feat(client): drop retry/backoff functionality

lukasmalkmus:


axiom/client.go
```diff
@@ -250,52 +248,12 @@ func (c *Client) NewRequest(ctx context.Context, method, path string, body any)
...
-		resp = newResponse(httpResp)
+	//nolint:bodyclose // The response body is closed later down below.
+	httpResp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, err
 	}
+	resp := newResponse(httpResp)
 
 	defer func() {
 		if resp != nil {
```
Rambatino, line 259: Question: when wouldn't the resp be nil here? This defer is instantiated after the check for err, and if err == nil, isn't there by definition a resp? As far as I can see `newResponse(httpResp)` doesn't affect resp being nil.

Rambatino: Oh I think because in previous code, it came before the err check?

Rambatino: I was just wondering whether it was that that was impacting the linting. But i guess it's to do with the wrapping inside of another struct?

lukasmalkmus: Ah, thanks! Yeah, that's definitely code not needed anymore, now :) Cleaned it up and also get rid of the annoying linter comment.

axiom/client.go
```diff
@@ -250,63 +248,16 @@ func (c *Client) NewRequest(ctx context.Context, method, path string, body any)
...
-			resp = newResponse(httpResp)
-
-			// We should only retry in the case the status code is >= 500,
-			// anything below isn't worth retrying.
-			if code := resp.StatusCode; code >= 500 {
-				_, _ = io.Copy(io.Discard, resp.Body)
-				_ = resp.Body.Close()
-
-				// Reset the requests body, so it can be re-read.
-				if req.Body, err = req.GetBody(); err != nil {
```
lukasmalkmus, line 282: I think we should also drop any code that sets `GetBody`.

PR #297, opened 2024-04-03
chore(monitors): updated monitors to pass in resolvable flag which is required by the API

Rambatino:
Have removed `err = fmt.Errorf("unknown comparison %q", s)`, as UI will contain some operators not present in this library and it saves breaking the integration specs.


toppercodes (approved):

Rambatino:
@toppercodes 
> Won't this break the api for anyone using one of these comparisons?
I've reset it back, it's because I'm running them against axiomers and I should do it on the e2e org

PR #302, opened 2024-04-23
drop restrictions for api tokens only working on ingest and query paths

toppercodes:


axiom/client.go
```diff
@@ -198,10 +195,6 @@ func (c *Client) NewRequest(ctx context.Context, method, path string, body any)
 	}
 	endpoint := c.config.BaseURL().ResolveReference(rel)
 
-	if config.IsAPIToken(c.config.Token()) && !validOnlyAPITokenPaths.MatchString(endpoint.Path) {
```
bahlo, line 201: Can we remove that method altogether?

toppercodes: It's used by `IsValidToken`

lukasmalkmus (approved):
LGTM

PR #311, opened 2024-05-22
quick docs fix to the zap example

konradjniemiec:
This example doesn't compile—I get: `missing dataset name`

### DIFF
--- a/adapters/zap/README.md
+++ b/adapters/zap/README.md
@@ -40,6 +40,7 @@ import (
 // ...
 
 core, err := adapter.New(
+    adapter.SetDataset("AXIOM_DATASET"),
     adapter.SetClientOptions(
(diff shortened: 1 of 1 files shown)


lukasmalkmus:
Hi @konradjniemiec 👋 

Thanks for the contribution! What you're describing is actually explained a couple of lines before: https://github.com/axiomhq/axiom-go/pull/311/files#diff-65932b4d21b82ce597f2aa3c53c235560d9da11836ed720eb5ca8355d4678859R23-R25.

Do you think this is not very obvious and we should improve it? Like do just one code block or something similar? What about other code examples like the ones in [examples](examples) or [GoDoc](https://pkg.go.dev/github.com/axiomhq/axiom-go@v0.17.8/adapters/zap#example-package)?

PR #317, opened 2024-06-10
feat: Add AnnotationsService

bahlo:
This adds support for Annotations: https://axiom.co/docs/restapi/endpoints/getAnnotations

### DIFF
--- a/axiom/annotations.go
+++ b/axiom/annotations.go
@@ -15,7 +15,7 @@ type Annotation struct {
 	ID          string    `json:"id"`
 	Datasets    []string  `json:"datasets"`
 	Time        time.Time `json:"time"`
-	EndTime     time.Time `json:"end_time"`
+	EndTime     time.Time `json:"endTime"`
 	Title       string    `json:"title"`
 	Description string    `json:"description"`
 	URL         string    `json:"url"`
@@ -29,7 +29,7 @@ type AnnotationCreateRequest struct {
 	Type     string   `json:"type"`
 	// Optional:
 	Time        time.Time `json:"time,omitempty"`
-	EndTime     time.Time `json:"end_time,omitempty"`
+	EndTime     time.Time `json:"endTime,omitempty"`
 	Title       string    `json:"title,omitempty"`
 	Description string    `json:"description,omitempty"`
 	URL         string    `json:"url,omitempty"`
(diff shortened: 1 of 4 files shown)


axiom/annotations.go
```diff
@@ -0,0 +1,172 @@
...
+// AnnotationsFilter filters annotations on the [`List`] method.
+type AnnotationsFilter struct {
+	Datasets []string   `url:"datasets"`
+	Start    *time.Time `url:"start"`
+	End      *time.Time `url:"start"`
+}
+
+// List annotations.
+func (a *AnnotationsService) List(ctx context.Context, filter *AnnotationsFilter) ([]*Annotation, error) {
+	ctx, span := a.client.trace(ctx, "Annotations.List", trace.WithAttributes())
```
lukasmalkmus, line 102: Please add attributes to `trace.WithAttributes()` or drop it. Ideally, add them conditionally if `filter != nil`.

axiom/annotations.go
```diff
@@ -0,0 +1,172 @@
...
+	var res Annotation
+	if err := a.client.Call(ctx, http.MethodGet, path, nil, &res); err != nil {
+		return nil, spanError(span, err)
+	}
+
+	return &res, nil
+}
+
+func (a *AnnotationsService) Update(ctx context.Context, id string, annotation *AnnotationUpdateRequest) (*Annotation, error) {
+	ctx, span := a.client.trace(ctx, "Annotations.Update")
```
lukasmalkmus, line 139: Add annotation ID as attribute.

axiom/annotations.go
bahlo: Fixed in 94f6afd

axiom/annotations.go
bahlo: Fixed in 94f6afd

axiom/annotations.go
```diff
@@ -0,0 +1,187 @@
...
+
+	var res Annotation
+	if err := a.client.Call(ctx, http.MethodPost, a.basePath, annotation, &res); err != nil {
+		return nil, spanError(span, err)
+	}
+
+	return &res, nil
+}
+
+// AnnotationsFilter filters annotations on the [`List`] method.
```
lukasmalkmus, line 93:
```
[`List`]
```

This is odd syntax. To render this properly in GoDoc, please do `[AnnotationsService.List]`.

axiom/annotations.go
```diff
@@ -0,0 +1,187 @@
...
+type AnnotationsFilter struct {
+	Datasets []string   `url:"datasets"`
+	Start    *time.Time `url:"start"`
+	End      *time.Time `url:"start"`
```
lukasmalkmus, line 97: I don't think those should be pointers. Pointers to times are annoying to use from a users perspective.

bahlo: 84efb90

axiom/annotations.go
bahlo: 84efb90

axiom/annotations.go
bahlo: f6d0be8

axiom/annotations.go
bahlo: f6d0be8

lukasmalkmus (approved):

PR #320, opened 2024-06-19
Add support for token management

toppercodes:


axiom/tokens_integration_test.go
```diff
@@ -0,0 +1,91 @@
+//go:build integration
```
lukasmalkmus, line 1: We don't use build tags anymore.

Then again, I just noticed they have been reintroduced partly with monitors and notifiers. Will clean that up, as well.

axiom/tokens_integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+	// list.
+	tokens, err := s.client.Tokens.List(s.ctx)
+	s.Require().NoError(err)
+	s.Require().NotEmpty(tokens)
+
+	s.Contains(tokens, s.apiToken)
+
+	// Regenerate the token and make sure the new token is part of the list.
+	regeneratedToken, err := s.client.Tokens.Regenerate(s.ctx, s.apiToken.ID, axiom.RegenerateTokenRequest{
+		ExistingTokenExpiresAt: time.Now().Add(-time.Second),
```
lukasmalkmus, line 74: Why the negative offset?

axiom/tokens_integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+	// Teardown routines use their own context to avoid not being run at all
+	// when the suite gets cancelled or times out.
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+
+	err := s.client.Tokens.Delete(ctx, s.apiToken.ID)
+	s.NoError(err)
+
+	s.IntegrationTestSuite.TearDownSuite()
+}
```
lukasmalkmus, line 41: Please use `TearDownTest`.

axiom/tokens_integration_test.go
```diff
@@ -0,0 +1,91 @@
...
+		ExpiresAt: time.Now().Add(24 * time.Hour),
+		DatasetCapabilities: map[string]axiom.DatasetCapabilities{
+			"*": {Ingest: []string{"create"}}},
+		OrganisationCapabilities: axiom.OrganisationCapabilities{
+			Users: []string{"create", "read", "update", "delete"},
+		}})
+	s.Require().NoError(err)
+	s.Require().NotNil(createdToken)
+
+	s.apiToken = createdToken.AsAPIToken()
```
lukasmalkmus, line 55: If creation of the token fails for some reason, the test suit will call `TearDownSuite` which in turn calls `s.apiToken.ID` where `apiToken` is `nil`. This will cause a panic.

Please move the creation logic into `SetupTest`, just like the teardown/deletion lives in `TearDown...`. Check the integrations tests for the annotations or datasets service for reference.

axiom/tokens_test.go
```diff
@@ -0,0 +1,256 @@
...
+        "name": "test",
+        "orgCapabilities": {
+            "apiTokens": [
+                "create"
+            ]
+        }
+    }]`, tokenTime.UTC().Truncate(time.Second).Format(time.RFC3339))
+		assert.NoError(t, err)
+	}
+	client := setup(t, "/v2/tokens/api", hf)
```
lukasmalkmus, line 62: Mismatches the API reference given as a comment on the services implementation.

axiom/tokens_test.go
```diff
@@ -0,0 +1,256 @@
...
+        "name": "test",
+        "orgCapabilities": {
+            "apiTokens": [
+                "create"
+            ]
+        }
+    }`, tokenTime.UTC().Truncate(time.Second).Format(time.RFC3339))
+		assert.NoError(t, err)
+	}
+	client := setup(t, "/v2/tokens/api/test", hf)
```
lukasmalkmus, line 114: Mismatches the API reference given as a comment on the services implementation.

axiom/tokens_test.go
```diff
@@ -0,0 +1,256 @@
...
+        "orgCapabilities": {
+            "apiTokens": [
+                "create"
+            ]
+        },
+		"token":"test"
+    }`, tokenTime.UTC().Truncate(time.Second).Format(time.RFC3339))
+		assert.NoError(t, err)
+	}
+	client := setup(t, "/v2/tokens/api", hf)
```
lukasmalkmus, line 167: Mismatches the API reference given as a comment on the services implementation.

axiom/tokens_test.go
```diff
@@ -0,0 +1,256 @@
...
+        "orgCapabilities": {
+            "apiTokens": [
+                "create"
+            ]
+        },
+		"token":"test"
+    }`, tokenTime.Add(24*time.Hour).UTC().Truncate(time.Second).Format(time.RFC3339))
+		assert.NoError(t, err)
+	}
+	client := setup(t, "/v2/tokens/api/test/regenerate", hf)
```
lukasmalkmus, line 234: Mismatches the API reference given as a comment on the services implementation.

axiom/tokens_test.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func TestTokensService_Delete(t *testing.T) {
+	hf := func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodDelete, r.Method)
+
+		w.WriteHeader(http.StatusNoContent)
+	}
+
+	client := setup(t, "/v2/tokens/api/testID", hf)
```
lukasmalkmus, line 252: Mismatches the API reference given as a comment on the services implementation.

axiom/tokens_test.go
```diff
@@ -0,0 +1,256 @@
...
+	"fmt"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var tokenTime = time.Now()
```
lukasmalkmus, line 14: Please don't use `time.Now()`. The implications this introduces are visible throughout the rest of the file.

You have to `sprintf` your responses and deal with truncation and timezones. Just use `testhelper.MustTimeParse` are use static strings, e.g. https://github.com/axiomhq/axiom-go/blob/main/axiom/datasets_test.go#L250.

lukasmalkmus (changes requested):

axiom/client.go
```diff
@@ -133,6 +134,7 @@ func NewClient(options ...Option) (*Client, error) {
 	client.Monitors = &MonitorsService{client: client, basePath: "/v2/monitors"}
 	client.Notifiers = &NotifiersService{client: client, basePath: "/v2/notifiers"}
 	client.Annotations = &AnnotationsService{client: client, basePath: "/v2/annotations"}
+	client.Tokens = &TokensService{client: client, basePath: "/v2/tokens/api"}
```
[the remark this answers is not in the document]
toppercodes: `/tokens` will limit future expansion of different token types.
There is no plan as of now to support different token types, so if preferred we can remove

axiom/tokens.go
```diff
@@ -0,0 +1,186 @@
...
+
+type RegenerateTokenRequest struct {
+	// ExistingTokenExpiresAt is the time when the existing token will expire.
+	ExistingTokenExpiresAt time.Time `json:"existingTokenExpiresAt"`
+	// NewTokenExpiresAt is the time when the new token will expire.
+	NewTokenExpiresAt time.Time `json:"newTokenExpiresAt"`
+}
+
+// Axiom API Reference: /v2/tokens
+type TokensService service
```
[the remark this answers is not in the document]
toppercodes: Accidental. But left comment on above

lukasmalkmus:
Left some comments, most of them rather trivial.

One bigger change I wanna discuss is the package level capabilities API.

My feeling is that the capabilities should be typed (permissions?). E.g.:

```go
cap := map[string]DatasetCapabilities{
	"dataset": {
		Ingest: []Permission{PermisionCreate},
		Query:  []Permission{PermisionRead},
	},
}
```

becomes:

```go
cap := map[string]DatasetCapabilities{
	"dataset": {
		Ingest: []Permission{PermisionCreate},
		Query:  []Permission{PermisionRead},
	},
}
```

Maybe `Permission` -> `APIPermission` but you get the idea. I think a prefix for the constants is a **must** because their names are to generic (essentially just CRUD verbs).

For an implementation, see `UserRole` in `users.go`.

axiom/tokens.go
```diff
@@ -0,0 +1,186 @@
...
+		attribute.String("axiom.token_id", id),
+	))
+	defer span.End()
+
+	path, err := url.JoinPath(s.basePath, id, "/regenerate")
+	if err != nil {
+		return nil, spanError(span, err)
+	}
+
+	var res CreateTokenResponse
```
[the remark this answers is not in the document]
toppercodes: Regenerating is creating a new token, but copy/pasting the permissions of an existing token. So it really is the createTokenResponse

axiom/tokens.go
```diff
@@ -0,0 +1,186 @@
...
+	if err := s.client.Call(ctx, http.MethodDelete, path, nil, nil); err != nil {
+		return spanError(span, err)
+	}
+
+	return nil
+}
+
+func (t *CreateTokenResponse) AsAPIToken() *APIToken {
+	return &t.APIToken
+}
```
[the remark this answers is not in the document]
toppercodes: Not needed and removed

axiom/tokens_integration_test.go
toppercodes: No reason but to ensure that the newly existing token is deleted

axiom/tokens_test.go
toppercodes: Addressed on comment above

axiom/tokens_test.go
toppercodes: Addressed on comment above

axiom/tokens_test.go
toppercodes: Addressed on comment above

axiom/tokens_test.go
toppercodes: Addressed on comment above

axiom/tokens_integration_test.go
toppercodes: Thanks have moved it to SetupTest

axiom/tokens_test.go
toppercodes: Thanks! Made that change

axiom/client.go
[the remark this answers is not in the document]
lukasmalkmus: /cc @bahlo But I would prefer.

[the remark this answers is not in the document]
bahlo: Yes, pls yeet

axiom/tokens_integration_test.go
lukasmalkmus: Ah, got it. So the server will basically list by doing `SELECT * FROM api_tokens WHERE expiry >= NOW()` (pseudo code, obviously 😄)? So you're trying to make sure it is definitely excluded, even if not yet purged from the table?

I'm not sure it matters, tbh. The latency should be high enough to prevent this from happening, e.g. `List/Get` will happen far enough after `Regenerate`.

Have you observed it failing? If not, either delete this **or** add a `HINT(topper codes): ...` one-liner, please.

axiom/tokens.go
```diff
@@ -0,0 +1,239 @@
...
+	// Flows is the Flows capability and the actions that can be performed on them.
+	Flows []Action `json:"flows,omitempty"`
+	// Integrations is the Integrations capability and the actions that can be performed on them.
+	Integrations []Action `json:"integrations,omitempty"`
+	// Monitors is the Monitors capability and the actions that can be performed on them.
+	Monitors []Action `json:"monitors,omitempty"`
+	// Notifiers is the Notifiers capability and the actions that can be performed on them.
+	Notifiers []Action `json:"notifiers,omitempty"`
+	// Rbac is the Rbac capability and the actions that can be performed on them.
+	Rbac []Action `json:"rbac,omitempty"`
```
lukasmalkmus, line 112:
```suggestion
	RBAC []Action `json:"rbac,omitempty"`
```

Needs some changes, elsewhere, as well.

axiom/client.go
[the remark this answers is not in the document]
toppercodes: Done. Will need to update api

axiom/tokens_integration_test.go
toppercodes: I've not observed it so will remove it

axiom/tokens.go
```diff
@@ -0,0 +1,239 @@
...
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"time"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+//go:generate go run golang.org/x/tools/cmd/stringer -type=Action -linecomment -output=action_string.go
```
lukasmalkmus, line 14: One last favour: Please rename this file to `tokens_string.go`. I get your intention - to name it after the only type it contains - but naming it `tokens` groups it nicely with all its related files (and is resistant to new types being added in the future and the filename suddenly not making sense anymore 😄)

axiom/tokens.go
```diff
@@ -0,0 +1,239 @@
...
+// Action represents an action that can be performed on an Axiom resource.
+type Action uint8
+
+// All available [Action].
+const (
+	ActionCreate Action = iota // create
+	ActionRead                 // read
+	ActionUpdate               // update
+	ActionDelete               // delete
+)
```
lukasmalkmus, line 25: I think it makes sense to use an `emptyAction` here (instead of `iota + 1`) so `Action(0)` doesn't evaluate to `ActionCreate`.

Check https://github.com/axiomhq/axiom-go/blob/main/axiom/orgs.go#L20-L29. The `orgs_test.go` file also has a pattern for testing a set of constants.

axiom/tokens_integration_test.go
```diff
@@ -0,0 +1,96 @@
...
+	suite.Run(t, new(TokensTestSuite))
+}
+
+func (s *TokensTestSuite) SetupSuite() {
+	s.IntegrationTestSuite.SetupSuite()
+}
+
+func (s *TokensTestSuite) TearDownSuite() {
+	s.IntegrationTestSuite.TearDownSuite()
+}
```
lukasmalkmus, line 31: No need to add that if you're not doing any custom setup in there. Since you're inheriting `IntegrationTestSuite`, this effectively just overrides `IntegrationTestSuite.SetupSuite` and `IntegrationTestSuite.TearDownSuite` just to call them from those method overrides.

lukasmalkmus (changes requested):
Three mini fixes. But please do add the unit tests for `Action`, tho.

axiom/tokens_integration_test.go
toppercodes: Thanks

PR #326, opened 2024-07-23
feat(monitors): add type to monitors to distinguish between Threshold & MatchEvent

Rambatino:


axiom/monitors.go
```diff
@@ -67,15 +82,37 @@ func (c *Operator) UnmarshalJSON(b []byte) (err error) {
...
 	ID string `json:"id,omitempty"`
+	// Type sets the type of monitor. Defaults to Threshold
+	Type Type `json:"type"`
 	// AlertOnNoData indicates whether to alert on no data.
 	AlertOnNoData bool `json:"alertOnNoData"`
 	// NotifyByGroup tracks each none-time group independently
 	NotifyByGroup bool `json:"notifyByGroup"`
 	// Resolvable determines whether the events triggered by the monitor
-	// are resolvable. This has no effect on threshold monitors
+	// are resolvable. This has no affect on threshold monitors
```
ImLunaHey, line 115: nit: was this intended?

Rambatino: It was intended as I figured it was being used as a verb, but turns out, no.

Rambatino, line 115:
```suggestion
	// are resolvable. This has no effect on threshold monitors
```

axiom/monitors.go
```diff
@@ -28,6 +28,13 @@ const (
 	AboveOrEqual // AboveOrEqual
 )
 
+type Type uint8
```
bahlo, line 31: This is package global, pls name it `MonitorType`

axiom/monitors.go
```diff
@@ -47,6 +54,14 @@ func operatorFromString(s string) (c Operator, err error) {
 	return c, err
 }
 
+func typeFromString(s string) (c Type) {
+	switch s {
+	case MatchEvent.String():
+		return MatchEvent
+	}
```
bahlo, line 61: is this not missing the threshold type?

axiom/monitors.go
```diff
@@ -96,6 +133,10 @@ type Monitor struct {
 	// Threshold the query result is compared against, which evaluates if the
 	// monitor acts or not.
 	Threshold float64 `json:"threshold"`
+	// CreatedAt.
```
lukasmalkmus, line 136: Proper sentences please, even if it seems redundant.

axiom/monitors.go
```diff
@@ -96,6 +133,10 @@ type Monitor struct {
...
+	CreatedAt time.Time `json:"createdAt"`
+	// CreatedBy.
```
lukasmalkmus, line 138: Proper sentences please, even if it seems redundant.

axiom/monitors.go
```diff
@@ -67,9 +82,31 @@ func (c *Operator) UnmarshalJSON(b []byte) (err error) {
...
+
+	*c = typeFromString(s)
+
+	return nil
+}
+
 type Monitor struct {
 	// ID is the unique ID of the monitor.
 	ID string `json:"id,omitempty"`
+	// Type sets the type of monitor. Defaults to Threshold
```
lukasmalkmus, line 108:
```suggestion
	// Type sets the type of monitor. Defaults to [Threshold].
```

axiom/monitors.go
lukasmalkmus, line 31: Please add a comment to this type explaining what it is. See the Operator comment for reference.

lukasmalkmus, line 31: Add this type to the `go:generate` line above.

axiom/monitors.go
```diff
@@ -28,6 +28,13 @@ const (
 	AboveOrEqual // AboveOrEqual
 )
 
+type Type uint8
+
+const (
```
lukasmalkmus, line 33: Please add a comment, see `Operator` for reference.

axiom/monitors.go
```diff
@@ -67,9 +82,31 @@ func (c *Operator) UnmarshalJSON(b []byte) (err error) {
...
+func (c *Type) UnmarshalJSON(b []byte) (err error) {
+	var s string
+	if err = json.Unmarshal(b, &s); err != nil {
+		return err
+	}
+
+	*c = typeFromString(s)
+
+	return nil
+}
```
lukasmalkmus, line 103: Please group these methods closer to the type declaration. E.g. every type, its constants and the methods are a logical block of code, e.g. for `Operator`:

```go
// Operator represents a comparison operation for a monitor. A [Monitor] acts on
// the result of comparing a query result with a threshold.
type Operator uint8

// All available [Monitor] comparison [Operator]s.
const (
	emptyOperator Operator = iota //

	Below        // Below
	BelowOrEqual // BelowOrEqual
	Above        // Above
	AboveOrEqual // AboveOrEqual
)

func operatorFromString(s string) (c Operator, err error) {
	switch s {
	case emptyOperator.String():
		c = emptyOperator
	case Below.String():
		c = Below
	case BelowOrEqual.String():
		c = BelowOrEqual
	case Above.String():
		c = Above
	case AboveOrEqual.String():
		c = AboveOrEqual
	default:
		err = fmt.Errorf("unknown operator %q", s)
	}

	return c, err
}

// MarshalJSON implements [json.Marshaler]. It is in place to marshal the
// Operator to its string representation because that's what the server
// expects.
func (c Operator) MarshalJSON() ([]byte, error) {
	return json.Marshal(c.String())
}

// UnmarshalJSON implements [json.Unmarshaler]. It is in place to unmarshal the
// Operator from the string representation the server returns.
func (c *Operator) UnmarshalJSON(b []byte) (err error) {
	var s string
	if err = json.Unmarshal(b, &s); err != nil {
		return err
	}

	*c, err = operatorFromString(s)

	return err
}
```

axiom/monitors.go
```diff
@@ -28,6 +28,13 @@ const (
 	AboveOrEqual // AboveOrEqual
 )
 
+type Type uint8
+
+const (
+	Threshold  Type = iota // Threshold
```
lukasmalkmus, line 34: If `Threshold` is not the default, consider adding an `empty`/`unknown` value. Is it?

axiom/monitors_integration_test.go
```diff
@@ -118,3 +118,24 @@ func (s *MonitorsTestSuite) Test() {
...
+			Range:         time.Minute * 10,
+			Threshold:     5,
+			Type:          axiom.MatchEvent,
+		},
+	})
+	s.Require().NoError(err)
+	s.Require().NotNil(monitor)
+
+	s.Equal(monitor.Type.String(), "MatchEvent")
+}
```
lukasmalkmus, line 141: Does `testify` call methods no prefixed with `Test...`? I don't think so because this is missing teardown code and every invocation of this test should create a new monitor but not delete it. And I see no monitors in our integration testing env.

axiom/monitors.go
lukasmalkmus, line 34: Is `Threshold` the exact casing the API returns?

axiom/notifiers.go
```diff
@@ -19,6 +19,10 @@ type Notifier struct {
 	DisabledUntil time.Time `json:"disabledUntil"`
 	// Properties of the notifier.
 	Properties NotifierProperties `json:"properties"`
+	// CreatedAt.
```
lukasmalkmus, line 22: Sentences.

axiom/notifiers.go
```diff
@@ -19,6 +19,10 @@ type Notifier struct {
...
+	CreatedAt time.Time `json:"createdAt"`
+	// CreatedBy.
```
lukasmalkmus, line 24: Sentences.

axiom/monitors_test.go
```diff
@@ -92,6 +93,7 @@ func TestMonitorsService_Get(t *testing.T) {
 	require.NoError(t, err)
 
 	assert.Equal(t, exp, res)
+	assert.Equal(t, res.Type, Threshold)
```
lukasmalkmus, line 96: Just extend the `exp` monitor with the expected type instead of adding an assertion here.

And just for future reference, in testify, the expected value comes before the one to assert. E.g. it should be:

```go
assert.Equal(t, Threshold, res.Type)
```

lukasmalkmus (changes requested):

axiom/monitors_operator_string.go
lukasmalkmus, line 1: This file shouldn't have been renamed.

axiom/monitors_type_string.go
lukasmalkmus, line 1: This file should not have been created. We use one `_string.go` file per "namespace", e.g. `monitors`, `datasets`, etc. This can be fixed by adjusting the `go:generate` comment accordingly.

axiom/monitors.go
Rambatino: It was defaulting to Threshold. Have refactored

axiom/monitors.go
```diff
@@ -67,9 +67,49 @@ func (c *Operator) UnmarshalJSON(b []byte) (err error) {
 	return err
 }
 
+// MonitorType represents the type of the monitor.
+type MonitorType uint8
+
+// All available [Monitor] [Types]s.
```
lukasmalkmus, line 73:
```suggestion
// All available [MonitorType]s.
```

axiom/monitors.go
```diff
@@ -67,9 +67,49 @@ func (c *Operator) UnmarshalJSON(b []byte) (err error) {
...
+	MonitorTypeThreshold  MonitorType = iota // Threshold
+	MonitorTypeMatchEvent                    // MatchEvent
+)
+
+func typeFromString(s string) (c MonitorType) {
+	switch s {
+	case MonitorTypeMatchEvent.String():
+		return MonitorTypeMatchEvent
+	default:
+		return MonitorTypeThreshold
```
lukasmalkmus, line 84: Keep in mind that this has the side effect of defaulting to `Threshold` if the API starts returning anything that's not `MatchEvent`. E.g. if you introduce a new type.

Not a blocker, tho.

axiom/monitors.go
```diff
@@ -96,6 +136,10 @@ type Monitor struct {
 	// Threshold the query result is compared against, which evaluates if the
 	// monitor acts or not.
 	Threshold float64 `json:"threshold"`
+	// CreatedAt is the time when the monitor was created.
+	CreatedAt time.Time `json:"createdAt"`
+	// CreatedBy is the user who created the monitor.
```
lukasmalkmus, line 141: Sorry, should have been more specific: Name or ID? Please put that in the GoDoc. E.g. `name of the user who ...` or `id of the user who ...`. Same for the other occurrence, please.

axiom/notifiers.go
```diff
@@ -19,6 +19,10 @@ type Notifier struct {
 	DisabledUntil time.Time `json:"disabledUntil"`
 	// Properties of the notifier.
 	Properties NotifierProperties `json:"properties"`
+	// CreatedAt is the time when the notifier was created.
+	CreatedAt time.Time `json:"createdAt"`
+	// CreatedBy is the user who created the notifier.
```
lukasmalkmus, line 24: Here, see comment above.

axiom/monitors.go
Rambatino: It's a fair point, this assumption does hold for now, but it won't in a months time. However, I don't want this to error when we introduce a new type. I think let's leave it for now, and I'll update it simultaneously when we do have a new type (as there will be new properties too).

lukasmalkmus (approved):

PR #333, opened 2024-08-19
chore(client): increase timeouts

### LINKED ISSUE #332: Increase timeouts to 2 Minutes
We need to increase the timeouts of axiom-go so that long queries don't error out.  Two minutes sounds like a reasonable amount.

@mhr3 mentioned that [The ResponseHeaderTimeout](https://github.com/axiomhq/axiom-go/blob/330d31b87ea80ad193b765c87a7d176be1d8ce4f/axiom/client.go#L74) need to be increased as well.

lukasmalkmus:
Some more sane timeout values.

Closes #332.


schehata (approved):
LGTM

PR #338, opened 2024-08-29
fix(notifiers): capitalising notifier keys

Rambatino:
The API responds with keys such as:

```
        "properties": {
            "ApiKey": "asdasdasdsadasdsad",
            "IsEU": true
        },
```

But when creating we lower camel case is. The backend is able to process them correctly, however they don't appear correctly in the UI:

image


ronoc (approved):

bahlo:
Our API is using camel case, not pascal case. Please fix on the backend ❤️

PR #343, opened 2024-10-03
Remove the pagerduty token field

toppercodes:
The pagerduty token is not required for out pagerduty notifier.

We're removing the need for the token from the api


lukasmalkmus (approved):
LGTM

PR #349, opened 2024-12-30
Idiomatic zerolog adapter readme

418Coffee:
The readme in the zerolog adapter is not quite idiomatic. It does not [separate logs by environment](https://axiom.co/docs/reference/datasets#separate-by-environment). It uses a non-existing function `WithDatasetName` instead of `adapter.SetDataset`. This PR addresses these issues.


lukasmalkmus (approved):
Very nice find!

Thanks for the contribution!

PR #350, opened 2025-01-10
feat: virtual fields to client

Rambatino:
Added virtual fields and tests to the client in preparation for: https://github.com/axiomhq/axiom/pull/12127 landing on production

### DIFF
--- a/axiom/vfields.go
+++ b/axiom/vfields.go
@@ -91,6 +91,8 @@ func (s *VirtualFieldsService) Create(ctx context.Context, req VirtualField) (*V
 func (s *VirtualFieldsService) Update(ctx context.Context, id string, req VirtualField) (*VirtualFieldWithID, error) {
 	ctx, span := s.client.trace(ctx, "VirtualFields.Update", trace.WithAttributes(
 		attribute.String("axiom.virtual_field_id", id),
+		attribute.String("axiom.param.dataset", req.Dataset),
+		attribute.String("axiom.param.name", req.Name),
 	))
 	defer span.End()
 
--- a/axiom/vfields.go
+++ b/axiom/vfields.go
@@ -36,7 +36,7 @@ type VirtualFieldsService service
 // List all virtual fields for a given dataset.
 func (s *VirtualFieldsService) List(ctx context.Context, dataset string) ([]*VirtualFieldWithID, error) {
 	ctx, span := s.client.trace(ctx, "VirtualFields.List", trace.WithAttributes(
-		attribute.String("axiom.param.dataset", dataset),
+		attribute.String("axiom.dataset_id", dataset),
 	))
 	defer span.End()
 
--- a/axiom/vfields.go
+++ b/axiom/vfields.go
@@ -40,11 +40,17 @@ func (s *VirtualFieldsService) List(ctx context.Context, dataset string) ([]*Vir
 	))
 	defer span.End()
 
-	params := url.Values{}
-	params.Set("dataset", dataset)
+	path, err := AddURLOptions(s.basePath, struct {
+		Dataset string `url:"dataset"`
+	}{
+		Dataset: dataset,
+	})
+	if err != nil {
+		return nil, spanError(span, err)
+	}
 
 	var res []*VirtualFieldWithID
-	if err := s.client.Call(ctx, http.MethodGet, s.basePath+"?"+params.Encode(), nil, &res); err != nil {
+	if err := s.client.Call(ctx, http.MethodGet, path, nil, &res); err != nil {
 		return nil, spanError(span, err)
 	}
 
--- a/axiom/vfields.go
+++ b/axiom/vfields.go
@@ -1,0 +1,127 @@
+package axiom
+
+import (
+	"context"
+	"net/http"
+	"net/url"
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+type VirtualField struct {
+	// Dataset is the dataset to which the virtual field belongs.
+	Dataset string `json:"dataset"`
+	// Name is the name of the virtual field.
+	Name string `json:"name"`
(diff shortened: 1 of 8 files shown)


axiom/vfields.go
```diff
@@ -0,0 +1,127 @@
...
+	))
+	defer span.End()
+
+	params := url.Values{}
+	params.Set("dataset", dataset)
+
+	var res []*VirtualFieldWithID
+	if err := s.client.Call(ctx, http.MethodGet, s.basePath+"?"+params.Encode(), nil, &res); err != nil {
+		return nil, spanError(span, err)
+	}
```
lukasmalkmus, line 49: Please use `AddURLOptions`: https://github.com/axiomhq/axiom-go/blob/main/axiom/datasets.go#L291-L296.

Your best bet is to use an anonymous struct to make this quick and easy, e.g.:

```go
_ = AddURLOptions(s.basePath, struct {
	Dataset string  `url:"dataset"`
}{
	Dataset: dataset,
})
```

axiom/vfields.go
```diff
@@ -0,0 +1,127 @@
...
+	ID string `json:"id"`
+}
+
+// Axiom API Reference: /v2/vfields
+type VirtualFieldsService service
+
+// List all virtual fields for a given dataset.
+func (s *VirtualFieldsService) List(ctx context.Context, dataset string) ([]*VirtualFieldWithID, error) {
+	ctx, span := s.client.trace(ctx, "VirtualFields.List", trace.WithAttributes(
+		attribute.String("axiom.param.dataset", dataset),
```
lukasmalkmus, line 39:
```suggestion
		attribute.String("axiom.dataset_id", dataset),
```

axiom/vfields.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, spanError(span, err)
+	}
+
+	return &res, nil
+}
+
+// Update the virtual field identified by the given id with the given properties.
+func (s *VirtualFieldsService) Update(ctx context.Context, id string, req VirtualField) (*VirtualFieldWithID, error) {
+	ctx, span := s.client.trace(ctx, "VirtualFields.Update", trace.WithAttributes(
+		attribute.String("axiom.virtual_field_id", id),
```
lukasmalkmus, line 93: Ideally, you wanna include the same params as in the `Create` method:

```suggestion
		attribute.String("axiom.virtual_field_id", id),
		attribute.String("axiom.param.dataset", req.Dataset),
		attribute.String("axiom.param.name", req.Name),
```

axiom/vfields_integration_test.go
```diff
@@ -0,0 +1,121 @@
...
+func (s *VirtualFieldsTestSuite) TearDownSuite() {
+	s.IntegrationTestSuite.TearDownSuite()
+}
+
+func (s *VirtualFieldsTestSuite) SetupTest() {
+	s.IntegrationTestSuite.SetupTest()
+
+	var err error
+	s.vfield, err = s.client.VirtualFields.Create(s.ctx, axiom.VirtualField{
+		Dataset:    "test-dataset",
```
lukasmalkmus, line 39: I assume the dataset doesn't actually need to exist?

From a users standpoint, not sure if that is a nice API 😅

axiom/vfields_integration_test.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (s *VirtualFieldsTestSuite) TestListVirtualFields() {
+	// List all virtual fields for the dataset and ensure the created field is part of the list.
+	vfields, err := s.client.VirtualFields.List(s.ctx, "test-dataset")
+	s.Require().NoError(err)
+	s.Require().NotEmpty(vfields)
+
+	s.Contains(vfields, s.vfield)
+}
```
lukasmalkmus, line 121: Looks like a copy of https://github.com/axiomhq/axiom-go/pull/350/files#diff-787497e8a6d9a01acf4c10ae02edf38456377b5487d92d7484128f0ee47fcc42R80-R85?

axiom/vfields_integration_test.go
```diff
@@ -0,0 +1,121 @@
...
+	s.Equal(vfield, fetchedField)
+
+	// Delete the virtual field.
+	err = s.client.VirtualFields.Delete(s.ctx, vfield.ID)
+	s.Require().NoError(err)
+
+	// Ensure the virtual field no longer exists.
+	_, err = s.client.VirtualFields.Get(s.ctx, vfield.ID)
+	s.Error(err)
+}
```
lukasmalkmus, line 112: Not sure what this test does. Different style but it tests all the same things as the `Test` method, no? Also be aware that if something goes after the create but before the delete, the field will be dangling in the integration test env and probably impact future runs of the test suite. Not ideal.

lukasmalkmus (dismissed):
Some nitpicks, good otherwise 👍

axiom/vfields_integration_test.go
Rambatino: It will return error if dataset doesn't exist

axiom/vfields_integration_test.go
Rambatino: Create vs Update does some deletion too. I'll combine.

axiom/vfields.go
```diff
@@ -0,0 +1,136 @@
...
+		attribute.String("axiom.dataset_id", dataset),
+	))
+	defer span.End()
+
+	path, err := AddURLOptions(s.basePath, struct {
+		Dataset string `url:"dataset"`
+	}{
+		Dataset: dataset,
+	})
+
```
lukasmalkmus, line 48:
```suggestion
```

axiom/vfields_integration_test.go
lukasmalkmus: But in that case, this test will error, no? Where is `test-dataset` being created.

lukasmalkmus (approved):

PR #355, opened 2025-02-19
feat(monitors): added missing monitor fields to client lib

Rambatino:
New fields to match API spec

### DIFF
--- a/axiom/monitors.go
+++ b/axiom/monitors.go
@@ -75,14 +75,17 @@ type MonitorType uint8
 
 // All available [MonitorTypes]s.
 const (
-	MonitorTypeThreshold  MonitorType = iota // Threshold
-	MonitorTypeMatchEvent                    // MatchEvent
+	MonitorTypeThreshold        MonitorType = iota // Threshold
+	MonitorTypeMatchEvent                          // MatchEvent
+	MonitorTypeAnonalyDetection                    // AnomalyDetection
 )
 
 func typeFromString(s string) (c MonitorType) {
 	switch s {
 	case MonitorTypeMatchEvent.String():
 		return MonitorTypeMatchEvent
+	case MonitorTypeAnonalyDetection.String():
+		return MonitorTypeAnonalyDetection
 	default:
 		return MonitorTypeThreshold
 	}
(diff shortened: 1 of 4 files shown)


axiom/monitors.go
```diff
@@ -145,6 +148,20 @@ type Monitor struct {
 	Disabled bool `json:"disabled"`
 	// DisabledUntil is the time that the monitor will be disabled until.
 	DisabledUntil time.Time `json:"disabledUntil"`
+	// SecondDelay is the delay in seconds after the end time that the monitor runs.
+	SecondDelay int64 `json:"secondDelay,omitempty"`
```
lukasmalkmus, line 152: Can you use a `time.Duration`, here? See how it's done for `Interval`, it is specifically marshalled/unmarshalled. I think it is the same for `Range`.

Couple of notes:

1. Add a comment saying the lowest resolution is seconds (and please do me a favour and add the same comment to `Interval` explaining that the lowest resolution is minutes)
2. Why is it that `secondDelay` has a lowercase `s` but `IntervalMinutes` has an uppercase `I`?

axiom/monitors_test.go
```diff
@@ -215,3 +215,185 @@ func TestMonitorsService_Delete(t *testing.T) {
 	err := client.Monitors.Delete(context.Background(), "testID")
 	require.NoError(t, err)
 }
+
+func TestMonitorsService_AdvancedUsage(t *testing.T) {
```
lukasmalkmus, line 219: Any reason these tests can't be folded into the `CreateMonitor` test?

The unit tests are there to make sure all params/fields/etc. are correctly passed to the server and back and that all the marshallers, etc. are involved correctly. There is no harm in having a single test that just populates **all the fields** for a create monitor request, even if that might be invalid when testing against a real environment. But in unit tests, everything is mocked, anyways.

I would have expected to do the "advanced testing" and more behavioural oriented tests in the integration test suite for monitors.

axiom/monitors_integration_test.go
```diff
@@ -58,14 +58,19 @@ func (s *MonitorsTestSuite) SetupTest() {
...
-			Threshold:     1,
+			AlertOnNoData:                false,
+			APLQuery:                     fmt.Sprintf("['%s'] | summarize count() by bin_auto(_time)", s.datasetID),
+			Description:                  "A test monitor",
+			Interval:                     time.Minute,
+			Name:                         "Test Monitor",
+			Operator:                     axiom.BelowOrEqual,
+			Range:                        time.Minute * 5,
+			Threshold:                    1,
+			SecondDelay:                  10,
```
lukasmalkmus, line 69:
```suggestion
			SecondDelay:                  time.Second * 10,
```

lukasmalkmus: Maybe add an assertion on the created monitor to make sure it is really 10 minutes?

axiom/monitors_test.go
```diff
@@ -192,11 +208,12 @@ func TestMonitorsService_Update(t *testing.T) {
 		DisabledUntil: parseTimeOrPanic("2024-05-12T18:43:37Z"),
 		ID:            "testID",
 		Interval:      0,
-		Name:          "test",
+		Name:          "newTest",
 		NotifierIDs:   nil,
 		Operator:      Above,
 		Range:         time.Minute,
 		Threshold:     1,
+		SecondDelay:   10,
```
lukasmalkmus, line 216: Same as above.

axiom/monitors_integration_test.go
Rambatino: ten seconds?

lukasmalkmus (approved):

PR #365, opened 2025-04-02
fix: ci

lukasmalkmus:
1. Adds new dataset fields.
2. Make sure to protect against nil pointer dereferences when tearing down a failed test.


Rambatino (approved):
Looks good

PR #367, opened 2025-04-02
chore: upgrade dependencies

lukasmalkmus:
Properly document that the underlying HTTP client used by default is equipped with a transport that is OTel enabled, even is `SetNoTracing` is provided. Users are expected to set their own client if they wish to prevent this.


peterbourgon (approved):

peterbourgon:
Please fix the PR title 😇

PR #372, opened 2025-04-29
Add source on slog adapter

VictorRibeiroLima:
Added the "AddSource" option to the slog adapter to better replicate the behavior of the default handler.

Since the slog.Record.source() method is private, I had to replicate its functionality.

I also didn't want to mess with the original "setup" method on the test so i created my own "setupWithSource"

### DIFF
--- a/adapters/slog/slog.go
+++ b/adapters/slog/slog.go
@@ -70,7 +70,7 @@ func SetLevel(level slog.Leveler) Option {
 	}
 }
 
-// SetAddSource specifies whether to add the source of the log message to The
+// SetAddSource specifies whether to add the source of the log message to The Event.
 func SetAddSource(addSource bool) Option {
 	return func(h *Handler) error {
 		h.addSource = addSource
--- a/adapters/slog/slog.go
+++ b/adapters/slog/slog.go
@@ -71,9 +71,9 @@ func SetLevel(level slog.Leveler) Option {
 }
 
 // SetAddSource specifies whether to add the source of the log message to the event.
-func SetAddSource(addSource bool) Option {
+func SetAddSource() Option {
 	return func(h *Handler) error {
-		h.addSource = addSource
+		h.addSource = true
 		return nil
 	}
 }
--- a/adapters/slog/slog.go
+++ b/adapters/slog/slog.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"log/slog"
 	"os"
+	"runtime"
 	"slices"
 	"sync"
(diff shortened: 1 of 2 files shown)


adapters/slog/slog.go
```diff
@@ -69,6 +70,14 @@ func SetLevel(level slog.Leveler) Option {
 	}
 }
 
+// SetAddSource specifies whether to add the source of the log message to The Event.
```
lukasmalkmus, line 73:
```suggestion
// SetAddSource specifies whether to add the source of the log message to the event.
```

adapters/slog/slog.go
```diff
@@ -69,6 +70,14 @@ func SetLevel(level slog.Leveler) Option {
...
+func SetAddSource(addSource bool) Option {
+	return func(h *Handler) error {
+		h.addSource = addSource
+		return nil
+	}
+}
```
lukasmalkmus, line 79: I don't have an extremely strong opinion here and I also believe it comes down to personal preference, **but** to be [consistent with existing options that set a boolean](https://github.com/axiomhq/axiom-go/tree/main/axiom/client_options.go#L78-L94), I suggest to not take a parameter and just set `true`:

```suggestion
func SetAddSource() Option {
	return func(h *Handler) error {
		h.addSource = true
		return nil
	}
}
```

WDYT?

lukasmalkmus (commented):
Nice, thanks for the contribution!

Code looks clean 👍 

Just left a nitpick and some food for thought. Feel free to to address and than we should be able to get this in pretty quickly.

adapters/slog/slog.go
VictorRibeiroLima: My opinion on this is nil.
i just coded this method really quick without paying much attention for the others options that set a bool.

lukasmalkmus:
@VictorRibeiroLima You need to fix the test to reflect removing the parameter to `SetAddSource()` and CI should pass: `Error: adapters/slog/slog_test.go:203:17: too many arguments in call to SetAddSource`.

VictorRibeiroLima:
> @VictorRibeiroLima You need to fix the test to reflect removing the parameter to `SetAddSource()` and CI should pass: `Error: adapters/slog/slog_test.go:203:17: too many arguments in call to SetAddSource`.

My Bad, i guess thats what happens when you do does things on the middle of your lunch break

Fixed!

lukasmalkmus (approved):
LGTM!

Thanks for the contribution!

PR #392, opened 2025-10-20
adapters/zerolog: fix counter increment before level check

lukasmalkmus:
Fixes a bug where the counter was incremented before checking if the log level was enabled, causing premature batch flushes.


Rambatino (approved):
Looks good 👌

PR #412, opened 2026-03-02
fix: pool zstd writers to mitigate leaking memory

roessland:
IngestEvents creates a new zstd.NewWriter on every call, allocating ~4 MB of encoder state each time. With the slog adapter flushing every second, this causes 336 MB peak heap in 2.5 minutes of tiny log traffic, and 896 MB under 4 concurrent callers — leading to OOM kills in production.

Fix: use a sync.Pool of zstd.Encoder with SpeedFastest level, reusing writers via Reset(). This reduces peak heap from 336 MB to 2 MB.

Containers running latest version of axiom-go requires seem to require at least ~100 MB more memory due to zstd encoder state, even with these fixes.

Disclaimer: PR description and code 100% AI generated.

### DIFF
--- a/axiom/datasets.go
+++ b/axiom/datasets.go
@@ -38,14 +38,15 @@ var ErrUnknownContentEncoding = errors.New("unknown content encoding")
 // zstd.NewWriter call. Writers are created with SpeedFastest to further reduce
 // memory footprint. Each writer is Reset() before use, which is safe and
 // discards all internal state from prior uses.
-var zstdWriterPool = sync.Pool{
-	New: func() any {
-		w, err := zstd.NewWriter(nil, zstd.WithEncoderLevel(zstd.SpeedFastest))
-		if err != nil {
-			panic("zstd: failed to create writer: " + err.Error())
+var zstdWriterPool sync.Pool
+
+func getZstdWriter() (*zstd.Encoder, error) {
+	if v := zstdWriterPool.Get(); v != nil {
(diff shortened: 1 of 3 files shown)


roessland:
Thanks for the review! I'll clean it up a bit and remove all the AI slop.

I noticed one of our newer internal services had a huge increase in memory usage. This service has very little traffic so I wasn't expecting much memory usage, and we hadn't set up alarms yet. The only symptom was occasional 502s from the load balancer when all instances (or request instance) OOM-ed simultaneously. 

Service was configured with 512 MB memory. Bumped to 1024 BM but still saw OOM-ing, but now less frequently.

image

2026-02-27 14:00:00 LOCAL: Upgraded axiom-go version from v0.26 (I think?) to v0.30.

image

10:00 ish: Memory bump 512M -> 1G
Until 12:45: Using axiomhq/axiom-go.
Until 14:50: Using https://github.com/roessland/axiom-go/commit/18011dd4c2f7a79838591e42948068b3da1a7605
After 14:50: Also using https://github.com/roessland/axiom-go/commit/13ca2a62136bf8599a4c71ff10f128c874cf7c56

This was a bit worrying since we have other more important services also using axiom-go (but luckily hadn't upgraded version yet).

lukasmalkmus:
@roessland Sorry, I forgot to follow up on this. We went ahead and merged https://github.com/axiomhq/axiom-go/pull/413 a while ago. Do you still want to work on something in this PR?

PR #425, opened 2026-04-06
fix: reset request body on network error retries and preserve failed …

thecraftman:
The retry logic in Client.Do only reset the request body (via req.GetBody) after HTTP 5xx responses but not after network errors such as "io: read/write on closed pipe". Because Go's HTTP transport consumes the request body during sending, subsequent retry attempts reused the already-consumed io.Pipe reader and failed immediately — making the entire retry mechanism ineffective for transient network failures.

Additionally, IngestChannel returned immediately on any flush error, discarding the in-flight batch of up to 10,000 events. The adapter's background loop would restart IngestChannel with an empty batch, causing silent data loss on every transient failure.

Changes:
- Reset request body before retrying on network errors in Client.Do, matching the existing behavior for 5xx retries
- Preserve failed batches in IngestChannel by retrying on the next tick instead of returning immediately, giving up only after 3 consecutive failures
- Add TestClient_Do_Backoff_NetworkError verifying body reset works for network errors (simulates io.ErrClosedPipe with consumed body)

Fixes a customer-reported issue where Logrus/slog adapter logging would stop ingesting after "io: read/write on closed pipe" errors, causing gaps in dashboards. The root cause was Railway (hosting platform) closing idle HTTP/2 connections, combined with these two bugs preventing recovery.

### DIFF
--- a/axiom/client.go
+++ b/axiom/client.go
@@ -270,6 +270,17 @@ func (c *Client) Do(req *http.Request, v any) (*Response, error) {
 			case errors.Is(err, context.Canceled):
 				return backoff.Permanent(err)
 			case err != nil:
+				// Reset the request body so it can be re-read on
+				// retry. Without this, retries after network errors
+				// (e.g. "io: read/write on closed pipe") would reuse
+				// the already-consumed body and fail immediately.
+				if req.GetBody != nil {
+					if body, resetErr := req.GetBody(); resetErr != nil {
+						return backoff.Permanent(resetErr)
+					} else {
+						req.Body = body
+					}
+				}
 				return err
 			}
 			resp = newResponse(httpResp)
--- a/axiom/client.go
+++ b/axiom/client.go
@@ -275,11 +275,11 @@ func (c *Client) Do(req *http.Request, v any) (*Response, error) {
 				// (e.g. "io: read/write on closed pipe") would reuse
 				// the already-consumed body and fail immediately.
 				if req.GetBody != nil {
-					if body, resetErr := req.GetBody(); resetErr != nil {
+					body, resetErr := req.GetBody()
+					if resetErr != nil {
 						return backoff.Permanent(resetErr)
-					} else {
-						req.Body = body
 					}
+					req.Body = body
 				}
 				return err
 			}
--- a/axiom/client.go
+++ b/axiom/client.go
@@ -274,13 +274,11 @@ func (c *Client) Do(req *http.Request, v any) (*Response, error) {
 				// retry. Without this, retries after network errors
 				// (e.g. "io: read/write on closed pipe") would reuse
 				// the already-consumed body and fail immediately.
-				if req.GetBody != nil {
-					body, resetErr := req.GetBody()
-					if resetErr != nil {
-						return backoff.Permanent(resetErr)
-					}
-					req.Body = body
+				body, resetErr := req.GetBody()
+				if resetErr != nil {
+					return backoff.Permanent(resetErr)
 				}
+				req.Body = body
 				return err
 			}
 			resp = newResponse(httpResp)
--- a/axiom/datasets.go
+++ b/axiom/datasets.go
@@ -657,9 +657,8 @@ func (s *DatasetsService) IngestChannel(ctx context.Context, id string, events <
 		setIngestStatusOnSpan(span, ingestStatus)
 	}()
 
-	// Track consecutive flush failures. After maxConsecutiveErrors the
-	// method gives up and returns the error so the caller (e.g. the adapter
-	// background loop) can decide what to do.
+	// 3 attempts × up to ~10s Client.Do backoff ≈ 30s resilience window
+	// before we give up and hand the batch back to the adapter's outer loop.
 	const maxConsecutiveErrors = 3
 	var consecutiveErrors int
 
--- a/axiom/datasets.go
+++ b/axiom/datasets.go
@@ -657,6 +657,12 @@ func (s *DatasetsService) IngestChannel(ctx context.Context, id string, events <
 		setIngestStatusOnSpan(span, ingestStatus)
 	}()
 
+	// Track consecutive flush failures. After maxConsecutiveErrors the
+	// method gives up and returns the error so the caller (e.g. the adapter
+	// background loop) can decide what to do.
+	const maxConsecutiveErrors = 3
+	var consecutiveErrors int
+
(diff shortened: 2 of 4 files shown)


axiom/client.go
```diff
@@ -270,6 +270,17 @@ func (c *Client) Do(req *http.Request, v any) (*Response, error) {
 			case errors.Is(err, context.Canceled):
 				return backoff.Permanent(err)
 			case err != nil:
+				// Reset the request body so it can be re-read on
+				// retry. Without this, retries after network errors
+				// (e.g. "io: read/write on closed pipe") would reuse
+				// the already-consumed body and fail immediately.
+				if req.GetBody != nil {
```
lukasmalkmus: Nit: this `req.GetBody != nil` guard is redundant. We're already inside `if req.GetBody != nil && !c.noRetry` at line 259, so `GetBody` is guaranteed non-nil here. Not a problem, just unnecessary.

axiom/datasets.go
```diff
@@ -687,12 +693,26 @@ func (s *DatasetsService) IngestChannel(ctx context.Context, id string, events <
 
 			if len(batch) >= batchSize {
 				if err := flush(); err != nil {
-					return &ingestStatus, spanError(span, err)
+					consecutiveErrors++
+					span.RecordError(err)
+					if consecutiveErrors >= maxConsecutiveErrors {
+						return &ingestStatus, spanError(span, err)
+					}
+					// Batch is preserved for retry on next tick.
```
lukasmalkmus: Nit: "retry on next tick" is only accurate for the ticker-based path (line 713). In this batch-full path, the batch stays at/above `batchSize`, so the very next incoming event will re-trigger `flush()` immediately, not on the next tick. Consider something like "Batch is preserved; flush retried on next event or tick."

thecraftman:
@lukasmalkmus pls take a look here.

axiom/client.go
thecraftman: done

axiom/datasets.go
thecraftman: done

lukasmalkmus:
Thanks Tola, second pass as promised. Diagnosis is correct and both fixes are directionally right. Two small asks before merge, plus a list of things I'll pick up myself afterwards.

## Before merge

**1. One-line rationale comment on `maxConsecutiveErrors = 3`** at `axiom/datasets.go:663`. Something like:

```go
// 3 attempts × up to ~10s Client.Do backoff ≈ 30s resilience window
// before we give up and hand the batch back to the adapter's outer loop.
const maxConsecutiveErrors = 3
```

So the next person to touch this doesn't change it blindly.

**2. PR description reword.** "Preserve failed batches in IngestChannel" slightly oversells the batch-preservation fix, since the batch is still dropped on the 4th strike when `IngestChannel` returns and the adapter restarts with an empty one. Suggested replacement:

> Preserve in-flight batches in `IngestChannel` across transient flush failures by extending the retry budget to 3 consecutive attempts before handing the batch back to the caller. Reduces (but does not eliminate) data loss on transient network errors.

Feel free to tweak the wording, main ask is just not overselling what the batch-preservation fix actually guarantees.

**Optional nit** (fine to skip): in `TestClient_Do_Backoff_NetworkError`, returning raw `io.ErrClosedPipe` from the `roundTripFunc` is fine as a mechanism check, but in production the transport error comes back wrapped in `*url.Error` by `http.Client.Do`. A one-line comment noting "simulating what `http.Transport` surfaces" would make the intent explicit, or `&url.Error{Op: "Post", URL: req.URL.String(), Err: io.ErrClosedPipe}` for the real shape. Not worth blocking on.

## Follow-ups I'll take as axiom-go maintainer

Out of scope for this PR but I don't want to lose them, so flagging for visibility:

- **Companion test for the combined `Client.Do` + `IngestChannel` path on a transport-level error.** Your existing `TestDatasetsService_IngestChannel_FlushRetry` is a valid unit test of the IngestChannel-level retry in isolation (`noRetry = true` is correct for that scope), but the body-reset fix's `err != nil` branch at `client.go:272` isn't currently exercised via an `IngestEvents`-shaped request body. Happy to pair on it if you want to stretch into the retry flow, otherwise I'll add it in a follow-up.
- **Defend against HTTP/2 pooled-connection stickiness** on the first network-error retry in `Client.Do`, via `CloseIdleConnections()` on the underlying `*http.Transport` (type-assertion guarded so custom transports are left alone). This is arguably the actual root-cause fix, since we currently rely on Go's HTTP/2 pool to evict dead `ClientConn`s, which is a known sharp edge.
- **Observability when `IngestChannel` gives up.** Counter plus a log line including `len(batch)` on the drop path. No way to measure drop rate externally right now.
- **`DatasetsService.Ingest(ctx, id, r io.Reader, ...)` at `datasets.go:404`** silently has zero retries because `req.GetBody == nil` for arbitrary readers. Retry loop at `client.go:259` is gated on `GetBody != nil`. Same bug class as the body-reset fix, different code path.
- **Close-channel shutdown path at `datasets.go:687`** still drops the batch on flush error during `Close()`.
- **Integration harness with a hostile proxy in CI** so the next connection-layer regression fails a test instead of reaching users.

I'll file these as issues once this lands. Good work on the diagnosis and the two fixes.

thecraftman:
@lukasmalkmus tested the network events here in my dataset 

-----
Screenshot 2026-04-08 at 16 31 56


---------

Summary

- Fix 1: Reset request body on network error retries (client.go)

When Client.Do encounters a network error (e.g. io: read/write on closed pipe), the request body has already been consumed. Without resetting it via GetBody(), subsequent retries reuse the spent body and fail immediately. This fix resets the body before each retry attempt.

- Fix 2: Preserve in-flight batches across transient flush failures (datasets.go)

Preserve in-flight batches in IngestChannel across transient flush failures by extending the retry budget to 3 consecutive attempts before handing the batch back to the caller. Reduces (but does not eliminate) data loss on transient network errors.

Test plan
- Unit tests pass: TestClient_Do_Backoff_NetworkError, TestDatasetsService_IngestChannel_FlushRetry, TestDatasetsService_IngestEvents_Retry
- Live ingest verified against Axiom (IngestEvents and IngestChannel both succeeded with 0 failures)

lukasmalkmus (approved):
LGTM. Both fixes are correct, rationale comment is in place, CI green, live-tested. Follow-up issues filed: #426, #427, #428, #429, #430, #431.

PR #434, opened 2026-04-27
Monitors: Add support for `mplQuery` field

CeRiAl:


lukasmalkmus (commented):
Ideally we should be having an integration test for this, as well?

CeRiAl:
> Ideally we should be having an integration test for this, as well?
:+1: 

Yes, sure, adding that. I'm adding support for mpl-queries to terraform atm, that's why I'm doing the necessary changes here first.

axiom/monitors_integration_test.go
```diff
@@ -202,3 +202,52 @@ func (s *MonitorsTestSuite) TestCreateAnomalyDetectionMonitor() {
...
+			Name:          "Test MPL Monitor",
+			Operator:      axiom.BelowOrEqual,
+			Range:         time.Minute * 5,
+			Threshold:     1,
+		},
+	})
+	s.Require().NoError(err)
+	s.Require().NotNil(monitor)
+	s.Require().NotEmpty(monitor.ID)
+	s.True((monitor.APLQuery == expectedQuery) != (monitor.MPLQuery == expectedQuery))
```
lukasmalkmus: This is smart but a pain to read and if it goes wrong one has no idea which condition actually failed. Just do:

```go
s.Equal(expectedQuery, monitor.MPLQuery)
s.Empty(monitor.APLQuery)
```

axiom/monitors_integration_test.go
```diff
@@ -202,3 +202,52 @@ func (s *MonitorsTestSuite) TestCreateAnomalyDetectionMonitor() {
...
+		defer cancel()
+		err := s.client.Monitors.Delete(ctx, monitor.ID)
+		s.NoError(err)
+	}()
+
+	got, err := s.client.Monitors.Get(s.ctx, monitor.ID)
+	s.Require().NoError(err)
+	s.Require().NotNil(got)
+	s.Equal(monitor.ID, got.ID)
+	s.True((got.APLQuery == expectedQuery) != (got.MPLQuery == expectedQuery))
```
lukasmalkmus: Dito

lukasmalkmus (dismissed):
Looks good but please address these two oddities :D

SayedAlesawy (approved):

lukasmalkmus (approved):

PR #443, opened 2026-05-12
axiom: add integration test verifying header event labels are applied

lukasmalkmus:
Adds a focused integration test (`TestDatasetsTestSuite/TestIngestEventLabelsApplied`) that ingests an event with a header-supplied label via `ingest.SetEventLabel`, then queries the dataset and asserts both that the `region` label field is present on the returned event and that its value matches what the SDK shipped in the `X-Axiom-Event-Labels` header.

## Context

The existing `TestDatasetsTestSuite/Test` and `TestDatasetsTestSuite/TestIngestWithMapFields` fail deterministically on `main` (see nightly run [25706497889](https://github.com/axiomhq/axiom-go/actions/runs/25706497889)) at the assertion `s.EqualValues(ingested.Len()+22, ingestStatus.ProcessedBytes)`. The 22 bytes correspond to the JSON-marshaled `X-Axiom-Event-Labels` header. The backend now returns `ProcessedBytes == ingested.Len()` exactly, so those header bytes are no longer counted in the byte accounting.

From the failure alone it was unclear whether the regression was purely an accounting change in `ProcessedBytes` or whether labels stopped being applied to events entirely. This test isolates the labeling question from the accounting question.

## Result

The test passes locally against staging on `main`'s tip, confirming that **labels still land on events** and **only the `ProcessedBytes` accounting changed**. Pushing as a draft so CI runs the same verification against both staging and development deployments across the Go matrix, and so the test stays in-tree as regression coverage going forward.

Follow-up (not in this PR): update the existing `Test`/`TestIngestWithMapFields` assertions to no longer expect `ProcessedBytes` to include label bytes (or revert the server change, depending on intent).

### DIFF
--- a/axiom/datasets_integration_test.go
+++ b/axiom/datasets_integration_test.go
@@ -371,6 +371,70 @@ func (s *DatasetsTestSuite) Test() {
 	s.Require().NoError(err)
 }
 
+// TestIngestEventLabelsApplied verifies that header-supplied event labels (set
+// via [ingest.SetEventLabel] and shipped as the `X-Axiom-Event-Labels` header
+// by the SDK) actually land on ingested events and are queryable. A failure
+// here means labels are not being applied server-side, independent of any
+// `ProcessedBytes` accounting changes.
+func (s *DatasetsTestSuite) TestIngestEventLabelsApplied() {
+	now := time.Now().Truncate(time.Second)
+	events := []axiom.Event{
+		{
+			ingest.TimestampField: now,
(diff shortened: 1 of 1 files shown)


lukasmalkmus:
Closing — verification done. Header-supplied event labels still land on events; only ProcessedBytes accounting changed (axiomhq/axiom#17316). Will follow up with an SDK-side fix that updates the existing assertions instead of adding new ones.

PR #451, opened 2026-07-06
axiom: reduce flakiness in dataset and edge integration tests

KhaledHegazy222:
The integration suites run against a live, eventually-consistent deployment, and a few tests asserted on freshly-ingested data (or shared a fixed dataset name), so they failed intermittently regardless of the change under test — most recently seen breaking unrelated dependabot PRs.

`TestIngestWithMapFields` queried immediately after ingest and asserted an exact field/column count, but the second event's flattened columns didn't always propagate in time (row counts were already correct, so only the schema lagged). It now polls with `Eventually` until the full schema is queryable before asserting.

The edge tests get two robustness tweaks: the ingest->query propagation wait goes from 30s to 60s to absorb slower propagation under load, and each test now uses a unique dataset name so a single transient teardown-delete failure can't cascade into `entity exists` errors for every subsequent test in the suite.

No production code changes; tests only.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/axiom/datasets_integration_test.go
+++ b/axiom/datasets_integration_test.go
@@ -656,18 +656,32 @@ func (s *DatasetsTestSuite) TestIngestWithMapFields() {
 	// Ingest some more data.
 	ingestObjectDataFn(s, ingestDataMapFields2)
 
-	now := time.Now().Truncate(time.Second)
-	startTime := now.Add(-time.Minute)
-	endTime := now.Add(time.Minute)
-
-	// Run a simple APL query...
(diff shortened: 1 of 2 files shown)


lukasmalkmus (changes requested):
Looks good, can you just please not modify the dataset suffix. I think if we run into 409 here that means our cleanup mechanism doesn't work as intended or tests are indeed trying to create duplicate resources.

I'd rather fix that separately :)

But since this creates a lot of datasets, creating many unique ones and not cleaning them up will just run into dataset limits at some point.

lukasmalkmus (approved):

PR #465, opened 2026-09-11
axiom: remove legacy query support

lukasmalkmus:
Removes the deprecated `querylegacy` package, `Client.QueryLegacy` and `DatasetsService.QueryLegacy`.

**Breaking:** callers must move to `Query`. The legacy endpoint stays available through `Client.Call` at `/v1/datasets/{id}/query`.

Also removes the `querylegacy` example, its CI entries, four `generate` prerequisites in the Makefile, and the `SA1019` lint exclusion. The adapter integration helper verifies ingest with APL.


kamstrup (approved):
Onward and upwards

